PR #27, opened 2020-02-03
Initial UICollectionView integration with tracking system

Clstroud:
collection_focus_demo mov

### DIFF
--- a/EyeSpeak/AppDelegate.swift
+++ b/EyeSpeak/AppDelegate.swift
@@ -11,13 +11,15 @@ import UIKit
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
-    var window: UIWindow?
-
+    var window: HeadGazeWindow?
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
         application.isIdleTimerDisabled = true
-        CoreDataStack.loadPersistentContainer("Eyespeak")
+        let window = HeadGazeWindow(frame: UIScreen.main.bounds)
+        window.rootViewController = UIStoryboard(name: "Main", bundle: nil).instantiateInitialViewController()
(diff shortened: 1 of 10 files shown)


EyeSpeak/AppDelegate.swift
```diff
@@ -11,13 +11,15 @@ import UIKit
...
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
-    var window: UIWindow?
-
+    var window: HeadGazeWindow?
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
         application.isIdleTimerDisabled = true
-        CoreDataStack.loadPersistentContainer("Eyespeak")
```
jrtb: 👋

EyeSpeak/UICollectionView+Gaze.swift
```diff
@@ -0,0 +1,43 @@
...
+    private func selectItem(for gaze: UIHeadGaze, with event: UIHeadGazeEvent?) {
+        let point = gaze.location(in: self)
+        guard let indexPath = indexPathForItem(at: point) else {
+            for path in indexPathsForSelectedItems ?? [] {
+                deselectItem(at: path, animated: true )
+            }
+            return
+        }
+        if !(indexPathsForSelectedItems ?? []).contains(indexPath) {
+            selectItem(at: indexPath, animated: true, scrollPosition: .init())
```
jrtb: does selectItem here mean hover over? assuming this is step one towards adding touch later.

Clstroud: This has been rejiggered to differentiate between highlight and selection 👍

EyeSpeak/AppDelegate.swift
Clstroud: I don't think we have any plans for Core Data long-term, so we can safely drop it for now. Once we get the rest of everything up to a point we're comfortable with, we can drop the legacy code (and Core Data along with it)

PR #31, opened 2020-02-04
Feature/category ui updates

jsmorgan42:
Changes included:
- Added updateSnapshot() function to refresh snapshot whenever new data needs to be displayed
- Updates selectedCategory variable & then updates snapshot to display new array of presets when a user taps a category cell
- Updates currentSpeechText & then updates snapshot to display new speech text when a user taps a preset cell


EyeSpeak/TextSelection/TextSelectionViewController.swift
```diff
@@ -258,6 +281,11 @@ class TextSelectionViewController: UICollectionViewController {
                 synthesizer.stopSpeaking(at: .immediate)
             }
             synthesizer.speak(utterance)
+            currentSpeechText = text
+            updateSnapshot()
+        case .category(let category):
+            selectedCategory = category
+            updateSnapshot()
```
Clstroud: Can updateSnapshot() be called in the didSet of these properties?

jsmorgan42: Updated to use didSet()

PR #35, opened 2020-02-05
Head distance transform scaling auto-tune

Clstroud:
This makes it feel a lot better no matter how near or far one's head is from lé screen

### DIFF
--- a/EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
+++ b/EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
@@ -1,10 +1,36 @@
 import Foundation
 import ARKit
 
+extension Notification.Name {
+    static let applicationDidAcquireGaze = Notification.Name("applicationDidAcquireGaze")
+    static let applicationDidLoseGaze = Notification.Name("applicationDidLoseGaze")
+}
+
+extension UIApplication {
+
+    fileprivate class Storage {
+
+        static fileprivate(set) var isGazeTrackingActive: Bool = false {
+            didSet {
+                guard oldValue != isGazeTrackingActive else { return }
+                if isGazeTrackingActive {
+                    NotificationCenter.default.post(name: .applicationDidAcquireGaze, object: nil)
+                } else {
+                    NotificationCenter.default.post(name: .applicationDidLoseGaze, object: nil)
+                }
+            }
+        }
+    }
+
+    var isGazeTrackingActive: Bool {
+        return Storage.isGazeTrackingActive
+    }
+}
+
+
 class UIHeadGazeViewController: UIViewController, ARSessionDelegate, ARSCNViewDelegate {
  
     private(set) var sceneview: ARSCNView?
-    private var faceAnchor: ARFaceAnchor?
 
     public var virtualCursorView: UIVirtualCursorView?
 
@@ -68,9 +94,10 @@ class UIHeadGazeViewController: UIViewController, ARSessionDelegate, ARSCNViewDe
     */
     private var lastGazeNDCLocation = CGPoint(x: 0.5, y: 0.5)
     func session(_ session: ARSession, didUpdate frame: ARFrame) {
+        guard let faceAnchor = faceAnchor else { return }
 
-        let lBlinkAmount = faceAnchor?.blendShapes[.eyeBlinkLeft]?.floatValue ?? 0.0
-        let rBlinkAmount = faceAnchor?.blendShapes[.eyeBlinkRight]?.floatValue ?? 0.0
+        let lBlinkAmount = faceAnchor.blendShapes[.eyeBlinkLeft]?.floatValue ?? 0.0
+        let rBlinkAmount = faceAnchor.blendShapes[.eyeBlinkRight]?.floatValue ?? 0.0
 
         let leftBlink = lBlinkAmount > 0.05
         let rightBlink = rBlinkAmount > 0.05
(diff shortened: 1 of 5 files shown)


EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
```diff
@@ -1,17 +1,43 @@
...
     private(set) var sceneview: ARSCNView?
-    private var faceAnchor: ARFaceAnchor?
 
     public var virtualCursorView: UIVirtualCursorView?
 
     private var previousGaze: UIHeadGaze? //inherent from UIHeadGazeCallback
     private var cursorPositionInterpolator = LowPassInterpolator<CGPoint>(filterFactor: 0.2, initialValue: .zero)
     private var ndcSmoothingInterpolator = LowPassInterpolator<SIMD2<Float>>(filterFactor: 0.05, initialValue: .zero)
-    
+    private var computedScale: CGFloat = 6
```
jrtb: 🧙‍♀️✨

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
```diff
@@ -1,17 +1,43 @@
...
 import ARKit
 
+extension Notification.Name {
+    static let applicationDidAcquireGaze = Notification.Name("applicationDidAcquireGaze")
+    static let applicationDidLoseGaze = Notification.Name("applicationDidLoseGaze")
+}
+
+extension UIApplication {
+
+    fileprivate class Storage {
```
jrtb: Extending UIApplication for storage feels a little odd to me. Is this a thing others use or expect to see? Or would it be better to make it its own thing?

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
```diff
@@ -1,17 +1,43 @@
 import Foundation
 import ARKit
 
+extension Notification.Name {
```
jrtb: Is there a pattern other than notifications we could use for this to prevent the proliferation of spaghetti notifications? i.e. would it be worth the time to architect how the various components of the app should interact to see if some other pattern would be a better fit / or make life easier for devs in the future?

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
```diff
@@ -100,12 +127,28 @@ class UIHeadGazeViewController: UIViewController, ARSessionDelegate, ARSCNViewDe
...
 
     private var headNode: SCNNode?
-    private var headAnchor: ARFaceAnchor?
+    private var faceAnchor: ARFaceAnchor? {
+        didSet {
+            let oldTracked = oldValue?.isTracked ?? false
+            let newTracked = faceAnchor?.isTracked ?? false
+            if oldTracked && !newTracked {
+                DispatchQueue.main.async {
+                    UIApplication.Storage.isGazeTrackingActive = false
```
jrtb: we need to hop on the the main thread here because the object receiving this notification has to do UI things, correct? is it possible that this side-effect may not be super obvious to future devs and may cause problems?

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
pg8wood: Hoping after MVP this can be configurable in Settings 👷‍♂️

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
Clstroud: It's just a place to stick things privately inside the extension since extensions can't add new properties.

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
Clstroud: For cases like this, it's a fairly standard UIKit convention when there's a multitude of parties that are potentially interested in observing a state change. For this we are notifying when the global status of "is tracking active right now?" changes.

Here's an example of this pattern (also on UIApplication): 
Screen Shot 2020-02-05 at 10 15 02 AM

Clstroud: It is possible that we could investigate using something like Combine, but I'm not as well versed in it at this point

jrtb: If we're being intentional about the pattern we're choosing and it's a thing folks have seen / expect to see, I'm 100% on-board 👊 Thanks for reflecting on this!!

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
Clstroud: External callers can not modify the state of `UIApplication.Storage.isGazeTrackingActive`, so I'm not super worried about future devs doing the wrong thing. Moving the dispatch to where postNotification is called would be reasonable, but I felt it was better to encapsulate the `UIApplication.Storage.isGazeTrackingActive = x` call itself since that is technically also a UIKit property and should be called on the main queue according to convention.

And to be clear: this is necessary because the ARKit callbacks happen on arbitrary background queues to reduce burden on the main queue.

EyeSpeak/HeadGazeLib/UIHeadGazeViewController.swift
Clstroud: FWIW, this is less of a magic number than it was before. I just escalated it up to a property that can be mutated dynamically as the face anchor moves.

PR #37, opened 2020-02-05
Fixing cell animation so it only happens for preset items

jsmorgan42:
Changes included
-Added guard to only transform attributes of preset items


EyeSpeak/UIComponents/PresetUICollectionViewFlowLayout.swift
```diff
@@ -10,14 +10,26 @@ import UIKit
 
 class PresetUICollectionViewCompositionalLayout: UICollectionViewCompositionalLayout {
     
+    var dataSource: UICollectionViewDiffableDataSource<TextSelectionViewController.Section, TextSelectionViewController.ItemWrapper>? {
```
Clstroud: Minor, inconsequential suggestion: this could be privatized since no consumers of this class need to access this property

PR #39, opened 2020-02-06
Demo readiness

pg8wood:


EyeSpeak/UICollectionView+Gaze.swift
```diff
@@ -49,14 +49,22 @@ extension UICollectionView {
     private func setItemHighlighted(_ highlighted: Bool, at indexPath: IndexPath) {
         if let cell = cellForItem(at: indexPath) {
             if highlighted {
+                guard let canHighlight = delegate?.collectionView?(self, shouldHighlightItemAt: indexPath), canHighlight else {
```
Clstroud: Can this be updated to use the same ?? style as the guards added below?

PR #54, opened 2020-02-17
Finishing prediction text

jsmorgan42:
Changes included:
- Updated predictions array to update when the user is typing new text
- Modifies current text string when a user selects a predictive text option


EyeSpeak/Features/Presets/PresetsViewController.swift
```diff
@@ -46,6 +46,7 @@ class PresetsViewController: UICollectionViewController, KeyboardSelectionDelega
     private var currentSpeechText: String = HintText.preset.rawValue {
         didSet {
             self.updateSnapshot()
```
jrtb:
```suggestion
            updateSnapshot()
```

EyeSpeak/Features/Presets/PresetsViewController.swift
```diff
@@ -46,6 +46,7 @@ class PresetsViewController: UICollectionViewController, KeyboardSelectionDelega
...
+            self.updatePredictions()
```
jrtb:
```suggestion
            updatePredictions()
```

EyeSpeak/Features/Presets/PresetsViewController.swift
```diff
@@ -104,9 +105,11 @@ class PresetsViewController: UICollectionViewController, KeyboardSelectionDelega
         }
     }
     
-    // TODO: Hook this up with the TextExpression predict() function when the user updates
-    // the text in the text field
-    let predictions: [TextPrediction] = []
+    private var predictions: [TextPrediction] = [] {
+        didSet {
+            self.updateSnapshot()
```
jrtb:
```suggestion
            updateSnapshot()
```

EyeSpeak/Utilities/TextExpression.swift
```diff
@@ -57,6 +57,10 @@ class TextExpression {
     func append(text: String) {
         self.value.append(text)
```
jrtb:
```suggestion
        value.append(text)
```

EyeSpeak/Utilities/TextExpression.swift
```diff
@@ -57,6 +57,10 @@ class TextExpression {
...
     }
+    
+    func updateString(text: String) {
+        self.value = text
```
jrtb:
```suggestion
        value = text
```

EyeSpeak/Features/Presets/PresetsViewController.swift
```diff
@@ -375,6 +390,20 @@ class PresetsViewController: UICollectionViewController, KeyboardSelectionDelega
...
-        currentSpeechText.append(character)
+        var characterCased = character
+        if !currentSpeechText.isEmpty && character != " "{
+            characterCased = character.lowercased()
+        }
+        
+        currentSpeechText.append(characterCased)
+    }
+    
+    private func updatePredictions() {
```
Clstroud: How would you feel about changing "Prediction" vocabulary to "Suggestion" since we're not actually predicting anything? It could keep the namespace more open for incorporating ML-based predictions in the future?

EyeSpeak/Utilities/TextExpression.swift
```diff
@@ -57,6 +57,10 @@ class TextExpression {
...
+    }
 
     func clear() {
         self.value = ""
```
jrtb:
```suggestion
        value = ""
```

EyeSpeak/Utilities/TextExpression.swift
```diff
@@ -57,6 +57,10 @@ class TextExpression {
     func append(text: String) {
         self.value.append(text)
     }
+    
+    func updateString(text: String) {
```
Clstroud: Given the `append()` above, can this action be named more descriptively? e.g. `replace(text: String)` ?

EyeSpeak/Features/Presets/PresetsViewController.swift
Clstroud: Do we need to call `updateSnapshot()` here, given `updatePredictions()` sets the `predictions` array and incurs `updateSnapshot()` anyway?

pg8wood:
> LGTM! Re: `self.` comments, I'm in the camp of not using `self.` unless you need to. My rational there is in the case where you need it inside closures, it's an indicator to think about whether you want a weak or strong reference. If you get in the habit of using it everywhere, you may not think about this when the compiler prompts you.


I like this :+1:

PR #60, opened 2020-02-19
Keyboard single key layout

jsmorgan42:
Changes included:
- Changed layout to show single key QWERTY layout instead of being grouped
- Implemented all keyboard function buttons (clear, backspace, space, speak)
- Created TextTransaction struct that abstracts all logic with handling the text field string. This includes adding highlighting/deleting logic that will highlight the full word when a suggestion option is selected & then delete the full word if the user then selects backspace. In addition, as a user types singular characters the last character is always highlighting, indicating that using the backspace button will only clear the last character.


EyeSpeak/Features/Presets/PresetUICollectionViewCompositionalLayout.swift
```diff
@@ -175,83 +175,18 @@ class PresetUICollectionViewCompositionalLayout: UICollectionViewCompositionalLa
...
+        
+        let keyboardContainerGroup = NSCollectionLayoutGroup.vertical(layoutSize: NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0), heightDimension: .fractionalHeight(1.0 / 2.0)), subitem: keyboardRowGroup, count: 3)
         
         let section = NSCollectionLayoutSection(group: keyboardContainerGroup)
         section.contentInsets = .init(top: 0, leading: 0, bottom: 0, trailing: 0)
         
         return section
     }
     
-    //                                 +-------+
```
jrtb: 👋

EyeSpeak/Features/Presets/Models/TextTransaction.swift
```diff
@@ -0,0 +1,103 @@
...
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+struct TextTransaction: CustomDebugStringConvertible {
+    
+    let text: String
+    let attrText: NSMutableAttributedString
```
Clstroud: Can this be fully spelled out? `attrText` -> `attributedText`. Not doing it as a suggestion here since that would break the build on its own.

EyeSpeak/Features/Presets/Models/TextTransaction.swift
```diff
@@ -0,0 +1,103 @@
...
+        case .none:
+            return TextTransaction(text: text, changeType: .lastCharacter)
+        }
+        return TextTransaction(text: newText, changeType: .lastCharacter)
+    }
+    
+    func appendingCharacter(with character: String) -> TextTransaction {
+        var characterCased = character
+        if !text.isEmpty && character != " "{
+            characterCased = character.lowercased()
```
Clstroud: Can this always be lowercased since the init will enforce capitalized casing?

EyeSpeak/Features/Presets/Models/TextTransaction.swift
```diff
@@ -0,0 +1,103 @@
...
+    
+    func insertingSuggestion(with suggestion: String) -> TextTransaction {
+        let newText: String
+        newText = NSString(string: text + " ").replacingCharacters(in: lastTokenRange, with: suggestion)
+        return TextTransaction(text: newText, changeType: .fullWord)
+    }
+    
+    // To keep track of when to delete the last word or the full word (after selecting a text suggestion) when pressing backspace
+    // on the keyboard.
+    enum ChangeType {
```
Clstroud: Can this enum be moved to the top of its enclosing type, for discoverability?

Clstroud: Also now that we have a deeper understanding, I propose we rename it to `TextTransaction.Intent` since `.none` doesn't make sense as a "type of change"

jsmorgan42: Changed

EyeSpeak/Features/Presets/Models/TextTransaction.swift
jsmorgan42: Yep, removed that code

EyeSpeak/Features/Presets/Models/TextTransaction.swift
jsmorgan42: Refactored 👍

PR #63, opened 2020-02-21
Preset pagination

pg8wood:
Allows pagination between a category of presets 

- Organizes preset items into a UIPageViewController for system-provided paging functionality

**Gaze**
RPReplay-Final1582243352

**Touch**
RPReplay_Final1582243812


EyeSpeak/Features/Presets/Models/TextPresets.swift
```diff
@@ -45,7 +45,19 @@ struct TextPresets {
...
+                     "I want another pillow.2",
+                     "I want another pillow.3",
+                     "I want another pillow.4",
+                     "I want another pillow.5",
+                     "I want another pillow.6",
+                     "I want another pillow.7",
+                     "I want another pillow.8",
+                     "I want another pillow.9",
+                     "I want another pillow.0",
+                     "I want another pillow.a",
```
jrtb: pillow.a -> brave choice. I would have gone with pillow.11

EyeSpeak/Features/Presets/Pagination/CategoryPaginationCollectionViewCell.swift
```diff
@@ -0,0 +1,20 @@
...
+//  CategoryPaginationCollectionViewCell.swift
+//  EyeSpeak
+//
+//  Created by Patrick Gatewood on 2/12/20.
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import UIKit
+
+class CategoryPaginationCollectionViewCell: PaginationContainerCollectionViewCell {
```
jrtb: 50/50 on how these two relate based on their names. If you all still like the names at this point, I'm along for the ride :)

EyeSpeak/Utilities/UIPageViewController+ManualPagination.swift
```diff
@@ -0,0 +1,42 @@
...
+extension UIPageViewController {
+    
+    /// A convenience method for setting the next or previous view controller to be displayed
+    func page(_ direction: UIPageViewController.NavigationDirection) {
+        switch direction {
+        case .forward:
+            pageForward()
+        case .reverse:
+            pageBackward()
+        @unknown default:
```
jrtb: I noticed this pattern before when switching on direction. Would it be better to just have the cases we know about currently and intentionally break if new ones are added in the future or do we have to have @unknown default for some reason?

Clstroud: @jrtb https://www.avanderlee.com/swift/unknown-default-enums-in-swift/ this is the way

jrtb: 🙇✌️🕊 thanks!! 🙂

EyeSpeak/Features/Presets/Pagination/CategoryPaginationCollectionViewCell.swift
pg8wood: Yeah I'm definitely not married to the names: We should use a naming scheme that allows us to separate:
- cell that contains the `UIPageViewController`
- collection view controller contained in the above type that contains category/preset item cells 
- cell that when selected will trigger the pagination action

Totally open to suggestions since I kicked a few names around already!

EyeSpeak/Features/Presets/PresetUICollectionViewCompositionalLayout.swift
```diff
@@ -13,6 +13,7 @@ class PresetUICollectionViewCompositionalLayout: UICollectionViewCompositionalLa
     // Height dimension of the product designs.
     // Intended for use in computing the fractional-size dimensions of collection layout items rather than hard-coding width/height values
     private static let totalHeight: CGFloat = 834.0
+    private static let totalWidth: CGFloat = 1130
```
Clstroud: Perhaps these would be more naturally expressed as a CGSize?

EyeSpeak/Features/Presets/Pagination/CategoryPaginationCollectionViewCell.swift
jrtb: Make Jesse decide ;) If nothing jumps out at you all after going through your excellent list of what these names should be, I'm good with moving on :)

EyeSpeak/Features/Presets/Pagination/PageIndicatorCollectionViewCell.swift
```diff
@@ -0,0 +1,25 @@
...
+
+import UIKit
+
+class PageIndicatorCollectionViewCell: VocableCollectionViewCell {
+    
+    @IBOutlet weak var pageLabel: UILabel!
+    
+    var pageInfo: String = "" {
+        didSet {
+            pageLabel.text = pageInfo
```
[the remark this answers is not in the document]
pg8wood: Nope!

PR #64, opened 2020-02-21
Introduce base for new phrase and category storage

Clstroud:
* (Re)introduces Core Data for storing phrases and categories
* The AppDelegate is responsible for purging existing non user-generated phrases/categories and reconstituting the latest versions at launch time
* Forthcoming work will bind the UI to this data model, but this seemed like a good point to PR and discuss


EyeSpeak/AppDelegate.swift
```diff
@@ -7,14 +7,19 @@
 //
 
 import UIKit
+import CoreData
```
jrtb: See PR #65 re: concerns about going straight to CoreData for storage. Is that the best fit? Seems like we need to bring in a lot of other complexity / scaffolding to get there. If we do want to go to CoreData, can we abstract out the storage bits into their own thing that has a simpler init/read/write contract? And then move the complexity out of AppDelegate and the VCs to clean up the call sites?

EyeSpeak/CoreData/NSManagedObject+Helpers.swift
```diff
@@ -0,0 +1,47 @@
...
+
+import CoreData
+
+protocol NSManagedObjectIdentifiable {
+    associatedtype IdentifierType
+}
+
+extension NSManagedObjectIdentifiable where Self: NSManagedObject {
+
+    static func fetchObject(in context: NSManagedObjectContext, matching identifier: IdentifierType) -> Self? {
```
pg8wood: Awesome

PR #65, opened 2020-02-21
Add saving of phrases

Clstroud:
* Enables use of heart button to save phrases

### DIFF
--- a/EyeSpeak/Features/Presets/PresetsViewController.swift
+++ b/EyeSpeak/Features/Presets/PresetsViewController.swift
@@ -8,6 +8,7 @@
 
 import UIKit
 import AVFoundation
+import CoreData
 
 class PresetsViewController: UICollectionViewController, PageIndicatorDelegate {
     
@@ -54,6 +55,7 @@ class PresetsViewController: UICollectionViewController, PageIndicatorDelegate {
     
     enum TopBarButton: String {
         case save
+        case unsave
         case toggleKeyboard
         case togglePreset
         case settings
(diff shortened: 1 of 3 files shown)


EyeSpeak/Features/Presets/PresetsViewController.swift
```diff
@@ -8,6 +8,7 @@
 
 import UIKit
 import AVFoundation
+import CoreData
```
jrtb: Could we extract storage into a separate thing outside this VC (separation of concerns / single responsibility principle)? Could we use UserDefaults instead of CoreData (potentially faster, simpler) for storage? I guess that would entail creating key:value pairs instead of saving the Phrase directly. But feels better to me?

pg8wood: +1 to moving `phraseIsSaved(:_)` to a shared location

EyeSpeak/Features/Presets/PresetsViewController.swift
```diff
@@ -322,13 +344,40 @@ class PresetsViewController: UICollectionViewController, PageIndicatorDelegate {
...
+                phrase.isUserGenerated = true
+                phrase.creationDate = Date()
+                phrase.lastSpokenDate = Date()
+                phrase.utterance = textTransaction.text
+                phrase.addToCategories(savedCategory)
+
+                do {
+                    try context.save()
+                } catch {
+                    assertionFailure("Failed to save user generated phrase: \(error)")
```
pg8wood: Not 100% on this, but similar to James' points would it be possible to wrap the saving and un-saving stuff up somewhere outside this ViewController?

pg8wood: Talked in person; we'll clean this up a bit in a separate PR

PR #68, opened 2020-02-25
Introduce Combine for observing selection state changes

pg8wood:
Currently, child view controllers either maintain their own version of the truth (e.g. selection state), or they ask their parent view controllers for the current state. 

This PR unifies the source of truth for those values that change over time, adds convenient access to and observation of these values.

Totally open to feedback on this, but as I was testing it out I found the implementations to be much cleaner than Notification Center and/or having a long chain of delegates


pg8wood:
> Looks cleaner for sure, although I don't know my Combine yet. I ❤️ the idea of being intentional about composition, sources of truth and how our VCs communicate with each other. Would love to sit-in next time this comes up!! Also interested in if it makes sense to use Combine and this pattern everywhere in the app and what that would take to do.

@jrtb great question! I think if we go this way the page indicator would be a perfect candidate to roll into Combine as well

EyeSpeak/Common/ItemSelection.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//  Created by Patrick Gatewood on 2/25/20.
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import Foundation
+import Combine
+import CoreData
+
+struct ItemSelection {
```
pg8wood: I don't love this name; it feels like it is similar to a `Store` in RxLand but I can't seem to think of something that I like

PR #74, opened 2020-03-02
Improvement/keyboard polish

jsmorgan42:
Changes included:
- Adding a function that handles grammatical rules around punctuation
- Rule 1: If a character is added after "." or "?" , the character should be capitalized 
- Rule 2: If a punctuation character is typed, the whitespace after the previous character should be removed (e.g. "However   ," -> "However,")
- Rule 3: If a character is typed after a punctuation symbol, the whitespace after the punctuation should be removed (e.g. "however,    that" -> "however, that")


pg8wood (approved):
LGTM! If we do any more advanced punctuation / sentence casing, we should think about looking into best practices and/or libraries designed around the more complex rules

EyeSpeak/Features/Presets/Models/TextTransaction.swift
```diff
@@ -87,6 +88,36 @@ struct TextTransaction: CustomDebugStringConvertible {
...
+                newText = trimmedText
+                newText += " "
+            }
+            
+            if lastCharacter == "." || lastCharacter == "?" {
+                char = char.uppercased()
+            }
+        }
+        
+        if punctuation.contains(char) {
```
thomasshealy: I think this could be `if punctation.contains(char) && newText.last == " " { newText = ...}`

EyeSpeak/Features/Presets/Models/TextTransaction.swift
```diff
@@ -87,6 +88,36 @@ struct TextTransaction: CustomDebugStringConvertible {
...
     }
     
+    func handleGrammar(with character: String) -> String {
+        var newText = text
+        var char = character.lowercased()
+        
+        let punctuation = ["'", ",", ".", "?"]
+        
+        let trimmedText = newText.trimmingCharacters(in: .whitespaces)
+        if let lastCharacter = trimmedText.last {
```
thomasshealy, line 98: I think this could be if `if let lastCharacter = trimmedText.last && punctation.contains(String(lastCharacter))`

PR #76, opened 2020-03-02
[Unknown Jira Number] - Out of Range Alert

thomasshealy:
I added a view which appears at the top of the screen whenever the user is out of head tracking range.  It disappears once the user is back within range.  

Screen Shot 2020-03-02 at 3 45 51 PM
Screen Shot 2020-03-02 at 3 45 41 PM
Screen Shot 2020-03-02 at 3 44 20 PM


EyeSpeak/Common/Views/TrackingContainerViewController.swift
```diff
@@ -10,10 +10,16 @@ import UIKit
 
 final class TrackingContainerViewController: UIViewController {
 
+    @IBOutlet weak var warningView: UIView!
```
jrtb: Would it be worth subclassing here to add warningView and its observers?

EyeSpeak/Common/Views/TrackingContainerViewController.swift
```diff
@@ -10,10 +10,16 @@ import UIKit
...
     @IBOutlet private var cursorView: UIVirtualCursorView!
     
     private var contentViewController: UIViewController!
     
+    override func viewDidLoad() {
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .applicationDidAcquireGaze, object: nil)
```
jrtb: Assuming the object responsible for creating these notifications is TBD/WIP?

jrtb (commented):
Excellent!!! 👊💰 One thing to think about and one question for you.

EyeSpeak/Common/Views/TrackingContainerViewController.swift
```diff
@@ -10,10 +10,16 @@ import UIKit
 
 final class TrackingContainerViewController: UIViewController {
 
+    @IBOutlet weak var warningView: UIView!
     @IBOutlet private var cursorView: UIVirtualCursorView!
     
     private var contentViewController: UIViewController!
     
+    override func viewDidLoad() {
```
Clstroud:
```suggestion
    override func viewDidLoad() {
        super.viewDidLoad()
```

EyeSpeak/Features/Settings/SettingsViewController.swift
```diff
@@ -12,9 +12,30 @@ import UIKit
...
     
     @IBOutlet var dismissButton: GazeableButton!
-    
+
     @IBAction func dismissSettings(_ sender: Any) {
         dismiss(animated: true, completion: nil)
     }
+    @IBOutlet weak var warningView: UIView!
+    
+    override func viewDidLoad() {
```
Clstroud:
```suggestion
    override func viewDidLoad() {
        super.viewDidLoad()
```

EyeSpeak/Features/Settings/SettingsViewController.swift
```diff
@@ -12,9 +12,30 @@ import UIKit
...
+        UIView.animate(withDuration: 0.5, animations: {
+            self.warningView.alpha = 0.0
+        })
+    }
+
+    @objc private func applicationDidLoseGaze(_ sender: Any?) {
+        UIView.animate(withDuration: 0.5, animations: {
+            self.warningView.alpha = 1.0
+        })
+    }
```
Clstroud: I'm not sure having this banner on a per-view controller level is going to scale well long-term. It already feels a bit duplicative in this form. Do you think we can have it be owned by the window itself so that one instance can be all-encompassing?

I think this would tie relatively cleanly into the existing mechanism that hides the cursor when gaze is lost

EyeSpeak/Supporting Files/Info.plist
```diff
@@ -24,6 +24,11 @@
 	<true/>
 	<key>NSCameraUsageDescription</key>
 	<string>Vocable uses ARKit head tracking to detect the user's face as seen in the device's TrueDepth front camera feed for the purpose of controlling an on-screen pointer, allowing users with limited mobility to interact with the app. Head orientation and position data is used expressly for controlling the on-screen pointer. That head tracking data is not stored or transmitted in any way. Vocable needs permission to use the camera for this purpose.</string>
+	<key>UIApplicationSceneManifest</key>
+	<dict>
+		<key>UIApplicationSupportsMultipleScenes</key>
+		<false/>
+	</dict>
```
Clstroud: Is this an intentional part of this PR? Looks correct, just seems unrelated to the feature so I wanted to ask :)

EyeSpeak/Features/Settings/Settings.storyboard
```diff
@@ -87,13 +87,50 @@
                                     <action selector="dismissSettings:" destination="gMn-gX-rfc" eventType="touchUpInside" id="B1M-NP-Phw"/>
                                 </connections>
                             </button>
+                            <view alpha="0.0" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="zNM-8r-lxH" userLabel="warningView">
```
pg8wood: I'm wondering if we can do this on the window-level so that we don't need to duplicate it here

pg8wood (commented):
Awesome work! The banner looks really good. I did notice that the banner appears if you disable head tracking in Settings, and I think we'll want to not show it in that case. Great stuff!

EyeSpeak/Features/Settings/Settings.storyboard
thomasshealy: Good call.

EyeSpeak/Supporting Files/Info.plist
thomasshealy: I think I had accidentally clicked this and committed, so I turned it back off.

EyeSpeak/Features/Settings/SettingsViewController.swift
thomasshealy: Good call.

EyeSpeak/Common/Views/TrackingContainerViewController.swift
thomasshealy: Sub classing turned out to be very difficult because of the `@objc` functions, I changed this so UIWindow owns the warning view which gets rid of the duplicate code.

EyeSpeak/Common/Views/TrackingContainerViewController.swift
thomasshealy: I'm not sure.

EyeSpeak/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -48,19 +51,55 @@ class HeadGazeWindow: UIWindow {
...
     required init?(coder: NSCoder) {
         super.init(coder: coder)
         commonInit()
     }
+    
+    private func initializeWarningView() {
+           warningView = UINib(nibName: "WarningView", bundle: .main).instantiate(withOwner: nil, options: nil).first as! UIView
+           
+           self.addSubview(warningView)
+
```
thomasshealy: Let me know if the deviceModel check is sketchy, it seems like it could be a decent work around to me.

EyeSpeak/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -48,19 +51,55 @@ class HeadGazeWindow: UIWindow {
...
+           let deviceModel = UIDevice().model.lowercased()
+           let width = deviceModel.contains("ipad") ? 425 : 350
+           
+           warningView.translatesAutoresizingMaskIntoConstraints = false
+           NSLayoutConstraint.activate([
+               warningView.topAnchor.constraint(equalTo: self.safeAreaLayoutGuide.topAnchor),
+               warningView.widthAnchor.constraint(equalToConstant: CGFloat(width)),
```
pg8wood: Let's talk about this in person. I think you're definitely on the right track thinking of things in terms of the size of the device, but I'd like to see if we can tackle it with constraints and/or size classes

Clstroud: For future reference, this would be the preferred API for checking the interface idiom and reacting `UIDevice.current.userInterfaceIdiom` 😄

pg8wood (approved):
Nice work! I like how responsive the alert is :+1:

PR #83, opened 2020-03-04
Additional tracking debug output

Clstroud:
In order to better diagnose what changes are needed for the iPhone, an additional runtime debug output view was needed. Additional output values can be added dynamically as tracking changes over time.
Screen Shot 2020-03-04 at 3 38 28 PM


stevefosterwta (approved):
🔍 👍

PR #87, opened 2020-03-05
Reset the categories PageViewController's pages when the size class changes

pg8wood:
Reset the categories PageViewController's pages when the size class changes


jrtb (approved):
👊💰

PR #89, opened 2020-03-05
Gaze cancellation support

Clstroud:
* Adds a means to cancel the current gaze to prevent repeated selections on momentarily selectable elements. 
* Fixes the infinite Settings <--> Presets navigation loop


pg8wood (approved):
Nice

PR #90, opened 2020-03-05
Settings for iPhone

thomasshealy:
Converted the settings screen to work with iPhones.

Screen Shot 2020-03-05 at 5 18 59 PM
Screen Shot 2020-03-05 at 5 18 52 PM

### DIFF
--- a/EyeSpeak/Common/Views/GazeableButton.swift
+++ b/EyeSpeak/Common/Views/GazeableButton.swift
@@ -28,7 +28,18 @@ class GazeableButton: UIButton {
     }
     
     var selectionFillColor: UIColor = .cellSelectionColor {
+        didSet {
+            updateContentViews()
+       }
+    }
     var buttonImageView = UIImageView()
+    private var imageSize: CGSize {
+        if traitCollection.horizontalSizeClass == .regular && traitCollection.verticalSizeClass == .regular {
+            return CGSize(width: 42, height: 42)
+        }
+        
+        return CGSize(width: 34, height: 34)
+    }
     
     override var isHighlighted: Bool {
         didSet {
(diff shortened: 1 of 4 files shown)


EyeSpeak/Features/Settings/SettingsViewController.swift
```diff
@@ -17,4 +17,12 @@ class SettingsViewController: UIViewController {
         dismiss(animated: true, completion: nil)
     }
     
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        dismissButton.buttonImage = UIImage(systemName: "xmark.circle")!
+    }
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+    }
```
stevefosterwta: Do we need this override? If so, I think we need to call super.

EyeSpeak/Common/Views/GazeableButton.swift
```diff
@@ -58,21 +66,24 @@ class GazeableButton: UIButton {
         layoutIfNeeded()
     }
     
-    private func sharedInit() {
+    func sharedInit() {
```
stevefosterwta: This should probably be left as `private`.

EyeSpeak/Features/Settings/SettingsCollectionViewController.swift
```diff
@@ -85,27 +85,91 @@ class SettingsCollectionViewController: UICollectionViewController, MFMailCompos
...
+        
+        let settingPageGroupSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0), heightDimension: .fractionalHeight(0.8))
+        let settingPageGroup = NSCollectionLayoutGroup.vertical(layoutSize: settingPageGroupSize, subitems: [settingsToggleGroup, settingsOptionsGroup, versionItem])
+        
+        let section = NSCollectionLayoutSection(group: settingPageGroup)
+        let layout = UICollectionViewCompositionalLayout(section: section)
+        return layout
+    }
+    
+    func compactHeight() -> UICollectionViewLayout {
```
pg8wood: I might name these functions something like `<size_class>Layout()` so that it's a bit more readable. Also if we're not using these outside of the class can they be `private`?

pg8wood (commented):
✅ LGTM pending the few small comments. Nice work dude!

jsmorgan42 (approved):
Awesome work!! 🚀

PR #104, opened 2020-03-06
Changing correct icon when on preset screen

jsmorgan42:
Changes includes:
- Now showing keyboard icon on preset screen for mode switching


stevefosterwta (approved):
👍

PR #110, opened 2020-03-09
Fixing layout for small iPhones

### LINKED ISSUE #108: Buttons are cut off in in landscape on handset

jsmorgan42:
Changes included:
- Fixing logic to only display compact portrait mode when having compact width & regular height


stevefosterwta (approved):
👍

PR #118, opened 2020-03-09
Custom Phrase Notification

thomasshealy:
Added a feature where a toast like message appears and fades out after a user has saved a custom phrase.  

Screen Shot 2020-03-09 at 5 37 53 PM
Screen Shot 2020-03-09 at 5 37 46 PM

### DIFF
--- a/EyeSpeak/HeadGazeLib/HeadGazeWindow.swift
+++ b/EyeSpeak/HeadGazeLib/HeadGazeWindow.swift
@@ -14,6 +14,7 @@ class HeadGazeWindow: UIWindow {
     let cursorView = UIVirtualCursorView()
     
     var warningView = UIView()
+    var phraseSavedView = UIView()
 
     private var trackingView: UIView?
     private var lastGaze: UIHeadGaze?
@@ -29,6 +30,7 @@ class HeadGazeWindow: UIWindow {
         cursorView.translatesAutoresizingMaskIntoConstraints = false
         self.addSubview(cursorView)
         initializeWarningView()
+        initializePhraseSaveView()
         
         NSLayoutConstraint.activate([
             cursorView.topAnchor.constraint(equalTo: self.topAnchor, constant: 0),
@@ -54,6 +56,7 @@ class HeadGazeWindow: UIWindow {
         super.addSubview(view)
         self.bringSubviewToFront(cursorView)
         self.bringSubviewToFront(warningView)
+        self.bringSubviewToFront(phraseSavedView)
     }
 
     required init?(coder: NSCoder) {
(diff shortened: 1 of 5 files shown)


stevefosterwta (approved):
👍

EyeSpeak/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -94,6 +112,15 @@ class HeadGazeWindow: UIWindow {
...
     }
     
+    @objc private func phraseSaved(_ sender: Any?) {
+        UIView.animate(withDuration: 1.5, animations: {
+            self.phraseSavedView.alpha = CGFloat(1.0)
+        })
+        UIView.animate(withDuration: 1.5, animations: {
+            self.phraseSavedView.alpha = CGFloat(0)
+        })
+    }
```
Clstroud, line 122: These are firing off at the same time and competing with each other. Can we either unify it as a keyframe animation or schedule the secondary animation in the completion block of the first?

EyeSpeak/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -74,11 +77,26 @@ class HeadGazeWindow: UIWindow {
                warningView.centerXAnchor.constraint(equalTo: self.centerXAnchor)
            ])
        }
+    
+    private func initializePhraseSaveView() {
+        phraseSavedView = UINib(nibName: "PhraseSavedView", bundle: .main).instantiate(withOwner: nil, options: nil).first as! UIView
+        self.addSubview(phraseSavedView)
+        let width = UIScreen.main.traitCollection.horizontalSizeClass == .compact ? 300 : 475
```
Clstroud, line 84: Can we compose these constraints in a way that doesn't require a hardcoded width value?

EyeSpeak/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -74,11 +77,26 @@ class HeadGazeWindow: UIWindow {
                warningView.centerXAnchor.constraint(equalTo: self.centerXAnchor)
            ])
        }
+    
+    private func initializePhraseSaveView() {
+        phraseSavedView = UINib(nibName: "PhraseSavedView", bundle: .main).instantiate(withOwner: nil, options: nil).first as! UIView
+        self.addSubview(phraseSavedView)
```
Clstroud, line 83: Can we add this to the view hierarchy lazily on the fly so it's not _always_ there, then remove it when we're done with it? I'd like to avoid any unnecessary compositing overhead with transparent views since we're trying to maintain 120hz as much as possible. The warning view may benefit from the same pattern.

PR #122, opened 2020-03-10
Possible fix when setting gazable alerts

stevefosterwta:


EyeSpeak/Common/Views/GazeableAlertController/GazeableAlertViewController.swift
```diff
@@ -48,6 +48,8 @@ final class GazeableAlertViewController: UIViewController {
     }
 
     public func setAlertTitle(_ title: String) {
+        guard isViewLoaded else { return }
+
         alertTitleLabel.text = title
```
Clstroud: This could be problematic if the title gets set and is not stored to be set later once `viewDidLoad` is called. We should store all the state we can and assign it in both the `didSet` and `viewDidLoad` to make sure both cases are covered

Clstroud (changes requested):

Clstroud (approved):
Looks good! 👍

PR #136, opened 2020-03-13
IBDesignable crash fix

Clstroud:
* Implements the workaround from https://openradar.appspot.com/47113341 to prevent the IB compiler from crashing incessantly
* Cleans up GazeableButton so it is once again touchable and less burdensome on the layout engine

### DIFF
--- a/Vocable/Extensions/UIColor+AppExtensions.swift
+++ b/Vocable/Extensions/UIColor+AppExtensions.swift
@@ -12,6 +12,7 @@ extension UIColor {
     
     // MARK: New Branded Colors
 
+    // Workaround for https://openradar.appspot.com/47113341. Prevent crashing the IBDesignables agent when using a color from the asset catalog.
     private convenience init?(safelyNamed name: String) {
         self.init(named: name,
                   in: Bundle(for: AppDelegate.self),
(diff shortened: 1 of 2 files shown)


Vocable/Extensions/UIColor+AppExtensions.swift
```diff
@@ -11,22 +11,28 @@ import UIKit
 extension UIColor {
     
     // MARK: New Branded Colors
-    
-    static let primaryColor = UIColor(named: "Primary")!
-    
-    static let defaultTextColor = UIColor(named: "DefaultFontColor")!
+
+    private convenience init?(safelyNamed name: String) {
```
pg8wood, line 16: Can we document the workaround with something like 

```suggestion
    // Workaround for https://openradar.appspot.com/47113341. Prevent crashing the IBDesignables agent when using a color from the asset catalog.
    private convenience init?(safelyNamed name: String) {
```

pg8wood (approved):
🚀

Vocable/Extensions/UIColor+AppExtensions.swift
Clstroud: This was a great idea you had. Thank you for the suggestion.

pg8wood: Always happy to help 😉

jsmorgan42 (approved):
🎸 ⚡️

PR #144, opened 2020-03-17
Fixes for My Sayings scrolling and deletion animations

### LINKED ISSUE #139: Animation bug when deleting my sayings cell
On the Edit My Sayings screen, if there is only one cell left on the next page & the user deletes the last cell on the first page, an animation bug occurs. See below.

ezgif com-video-to-gif

Clstroud:
Resolves #139 
* Fixes an issue where deleting the last item on a page would not scroll back to the previous page correctly
* Improves touch handling performance
* Prevents cell selection/highlight on My Sayings

### DIFF
--- a/Vocable/Common/CarouselGridCollectionViewController.swift
+++ b/Vocable/Common/CarouselGridCollectionViewController.swift
@@ -119,8 +119,11 @@ class CarouselGridLayout: UICollectionViewLayout {
     @Published
     var progress: CarouselGridPagingProgress?
 
+    private var lastInvalidatedSize: CGSize = .zero
+    private var lastInvalidatedVisiblePages = Set<Int>()
+    
     private var numberOfPages: Int {
-        guard let collectionView = collectionView else { return 1 }
+        guard let collectionView = collectionView, collectionView.window != nil else { return 1 }
         let pageCount = Int((Double(collectionView.numberOfItems(inSection: 0)) / Double(itemsPerPage)).rounded(.up))
         if pageCount != (progress?.pageCount ?? 0) {
             progress = (pageIndex: logicalPageIndex, pageCount: pageCount)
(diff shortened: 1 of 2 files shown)


Vocable/Common/CarouselGridCollectionViewController.swift
```diff
@@ -174,13 +177,17 @@ class CarouselGridLayout: UICollectionViewLayout {
...
+            return
+        }
 
         // Handles the case of the last item being deleted from the visible page
         if logicalPageIndex >= numberOfPages {
-            let rect = boundsRectForPageIndex(-1)
-            collectionView.scrollRectToVisible(rect, animated: animateIfNeeded)
-            logicalPageIndex -= 1
+            // Snap to the rightmost page before animating smoothly back to center.
+            collectionView.scrollRectToVisible(boundsRectForPageIndex(2), animated: false)
```
jsmorgan42, line 188: Why is this a hard-coded value of 2?

Vocable/Common/CarouselGridCollectionViewController.swift
```diff
@@ -197,6 +204,18 @@ class CarouselGridLayout: UICollectionViewLayout {
         return true
     }
 
+    private func visiblePages(forBounds bounds: CGRect) -> Set<Int> {
+        var pages = Set<Int>([0])
+        let tx = bounds.origin.x.distance(to: resetRect.origin.x)
```
jsmorgan42, line 209: What does the t represent in this variable name?

jsmorgan42 (approved):
LGTM, had a couple questions

Vocable/Common/CarouselGridCollectionViewController.swift
Clstroud: Because the there are 3 "physical" pages of the scrollview that we use to represent all potential logical pages. Indexed [0, 1, 2]. It constantly resets back to 1 so in this case we need to snap over to 2 and animate back to 1 to get the right effect.

Vocable/Common/CarouselGridCollectionViewController.swift
Clstroud: Translation. Just like the `.tx` and `.ty` instance properties of `CGAffineTransform`

stevefosterwta (approved):
👍

PR #148, opened 2020-03-18
ARKit Dependency Removal

### LINKED ISSUE #141: ARKit != TrueDepth camera in the UIRequiredDeviceCapabilities list
Users without ARKit supported devices are able to download and use the app in the App Store. Figure out if we want to support this and either change the requirement or allow it to work without head-tracking.

Clstroud:
Fixes #141 

* Removes the ARKit tracking view controller and virtual cursor from the view hierarchy when head tracking is unavailable or disabled
* Restructures toast notification and tracking error views so they are dynamically added to the view hierarchy
* Refactors Combine code to use a new property wrapper that wraps `UserDefaults` and also mimics `@Published`
* Removes ARKit dependency from Info.plist


thomasshealy (approved):
LGTM

thomasshealy (approved):
LGTM

PR #157, opened 2020-03-22
Selection Mode Screen

thomasshealy:
Built out the new selection mode screen which now contains the option to enable/disable head tracking.  This functionality has been moved from the main settings screen page. 

Screen Shot 2020-03-22 at 4 29 37 PM
Screen Shot 2020-03-22 at 4 29 33 PM
Screen Shot 2020-03-22 at 4 28 59 PM
Screen Shot 2020-03-22 at 4 29 07 PM


Vocable/Features/Settings/SelectionModeCollectionViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+        let headTrackingToggleItem = NSCollectionLayoutItem(layoutSize: NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0), heightDimension: .fractionalHeight(1.0)))
+        let headTrackingToggleGroupSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0), heightDimension: .fractionalHeight(1/5))
+        let headTrackingToggleGroup = NSCollectionLayoutGroup.vertical(layoutSize: headTrackingToggleGroupSize, subitems: [headTrackingToggleItem])
+        
+        let section = NSCollectionLayoutSection(group: headTrackingToggleGroup)
+        let layout = UICollectionViewCompositionalLayout(section: section)
+        return layout
+    }
+    
+    private func regularHeightWidthLayout() -> UICollectionViewLayout {
```
jsmorgan42: Using the name "defaultLayout" may make more sense in this case since it includes iPhones with compact height

jsmorgan42 (changes requested):
Can't find the line of code where it's changing this, but I noticed the switch in the cell looks disabled or user interaction is disabled (it's grayed out). Is there a way to make this toggle not grayed out? Otherwise looks great 🙂

Vocable/Features/Settings/SelectionModeCollectionViewController.swift
thomasshealy: I like the name change.  The grayed out switch seems to happen if head tracking is not supported, have you tried running it on a physical device with head tracking capabilities?

Clstroud: Yes, the graying out was part of #141. We don't want to be able to flip that switch if the device doesn't support it. I believe there is a note on the board for designing some footer text to describe why the switch is disabled, per iOS convention.

Vocable/Features/Settings/SelectionModeCollectionViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+    
+    func setupCollectionView() {
+        collectionView.backgroundColor = .collectionViewBackgroundColor
+        collectionView.register(UINib(nibName: "SettingsToggleCollectionViewCell", bundle: nil), forCellWithReuseIdentifier: "SettingsToggleCollectionViewCell")
+        
+        let layout = createLayout()
+        collectionView.collectionViewLayout = layout
+    }
+    
+    func createLayout() -> UICollectionViewLayout {
```
Clstroud:
```suggestion
    private func createLayout() -> UICollectionViewLayout {
```

Vocable/Features/Settings/SelectionModeCollectionViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+class SelectionModeCollectionViewController: UICollectionViewController {
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        setupCollectionView()
+    }
+
+    // MARK: UICollectionViewDataSource
+    
+    func setupCollectionView() {
```
Clstroud:
```suggestion
    private func setupCollectionView() {
```

Vocable/Features/Settings/SelectionModeCollectionViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+            collectionView.deselectItem(at: indexPath, animated: true)
+        }
+        
+        if AppConfig.isHeadTrackingEnabled {
+            let alertViewController = GazeableAlertViewController.make { AppConfig.isHeadTrackingEnabled.toggle() }
+            present(alertViewController, animated: true)
+            alertViewController.setAlertTitle("Turn off head tracking?")
+        } else {
+            AppConfig.isHeadTrackingEnabled.toggle()
+        }
```
Clstroud:
```suggestion
        let alertViewController = GazeableAlertViewController.make { AppConfig.isHeadTrackingEnabled.toggle() }
        present(alertViewController, animated: true)
        alertViewController.setAlertTitle("Turn off head tracking?")
```
The cell should not be selectable if head tracking is disabled. Check the implementation from the other view controller for the delegate methods to override for that. I believe highlight and selection were mitigated via delegate methods.

Vocable/Features/Settings/SettingsCollectionViewController.swift
```diff
@@ -227,6 +227,10 @@ class SettingsCollectionViewController: UICollectionViewController, MFMailCompos
             if let vc = self.storyboard?.instantiateViewController(identifier: "MySayings") {
                 show(vc, sender: nil)
             }
+        case .selectionMode:
+            if let vc = self.storyboard?.instantiateViewController(identifier: "SelectionMode") {
+                 show(vc, sender: nil)
+             }
```
Clstroud: Can this new screen be its own storyboard? Some of the existing Settings screens could probably be broken out, but since this is a new one I think it could lead the charge. Smaller storyboards == less conflicts later :)

Clstroud (changes requested):
This will seem like overkill at this stage, but I would really like to see this screen use diffable datasource. It will make life much easier as we add more to this screen in the coming days/weeks. Other than that, this is all good stuff 😄 I think the rest of the suggestions are relatively minor stuff.

Clstroud (approved):
💯 Nice job!

Vocable/Features/Settings/Settings.storyboard
```diff
@@ -248,7 +248,7 @@
                                     <segue destination="W98-u1-6GB" kind="embed" identifier="CarouselCollectionViewController" id="YT0-TB-czc"/>
                                 </connections>
                             </containerView>
-                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="xsL-XA-YD7" customClass="PaginationView" customModule="Vocable" customModuleProvider="target">
+                            <view contentMode="scaleToFill" ambiguous="YES" translatesAutoresizingMaskIntoConstraints="NO" id="xsL-XA-YD7" customClass="PaginationView" customModule="Vocable" customModuleProvider="target">
```
stevefosterwta, line 251: `ambiguous` might indicate there is a layout issue in `Settings.storyboard`

Vocable/Features/Settings/SettingsCollectionViewController.swift
```diff
@@ -303,4 +305,11 @@ class SettingsCollectionViewController: UICollectionViewController, MFMailCompos
         controller.dismiss(animated: true)
     }
     
+    private func presentSelectionModeViewController() {
+        let storyboard = UIStoryboard(name: "SelectionMode", bundle: nil)
+        let vc = storyboard.instantiateInitialViewController()!
+        vc.modalPresentationStyle = .overFullScreen
+        self.present(vc, animated: true, completion: nil)
```
stevefosterwta, line 327: I don't think `self.` isn't necessary here. You can also leave off `completion: nil` too.

jsmorgan42 (approved):

PR #160, opened 2020-03-23
feature/edit-saved-toast

thomasshealy:
Added a toast like message which gets displayed when  a user saves an edit to an existing phrase.

Screen Shot 2020-03-23 at 5 48 54 PM
Screen Shot 2020-03-23 at 5 49 51 PM


Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -11,6 +11,11 @@ import Combine
 import ARKit
 
 class HeadGazeWindow: UIWindow {
+    
+    private enum phraseSavedNibName: String {
+        case phraseSaved = "PhraseSavedView"
+        case editedPhraseSaved = "EditedPhraseSavedView"
+    }
```
Clstroud: Is there a way to make this use the existing mechanism by making what's in `develop` slightly more generic? I'm not sure we want to go down the path of having a separate xib for each potential instance of this toast, unless there's another aspect of this that I'm not seeing. It seems like we need some means to present a toast with configurable text and it would always be the same?

thomasshealy: Yeah, I'm thinking now that this could be an extension of UIView, I will look into it.

Vocable/Common/Views/PhraseSavedView.swift
```diff
@@ -10,7 +10,10 @@ import UIKit
 
 @IBDesignable
 class PhraseSavedView: BorderedView {
-
+
```
Clstroud, line 13:
```suggestion
```

Vocable/Common/Views/PhraseSavedView.swift
```diff
@@ -10,7 +10,10 @@ import UIKit
 
 @IBDesignable
 class PhraseSavedView: BorderedView {
-
+    
+    
+    @IBOutlet weak var alertLabel: UILabel!
```
Clstroud: Since this is an attribute of the view itself, we can probably get away with calling it a `titleLabel` and making it private. This would be akin to something like `UIButton` having a property called `buttonLabel`, which feels redundant.

Vocable/Common/Views/PhraseSavedView.swift
```diff
@@ -28,4 +31,8 @@ class PhraseSavedView: BorderedView {
         setContentHuggingPriority(.required, for: .vertical)
         setContentHuggingPriority(.required, for: .horizontal)
     }
+    
+    func setAlertText(text: String) {
+        alertLabel.text = text
+    }
```
Clstroud: Can this be a `var text: String` with a `didSet`? So for callers it would just be e.g. `toastView.text = "Saved to My Sayings"`?

Vocable/Common/Views/PhraseSavedView.swift
```diff
@@ -10,7 +10,10 @@ import UIKit
 
 @IBDesignable
 class PhraseSavedView: BorderedView {
```
Clstroud: Since this is no longer specific to saving phrases, can it be named something generic like `ToastView`?

Vocable/HeadGazeLib/HeadGazeWindow.swift
Clstroud: Is this still relevant in the latest revision?

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -55,7 +60,6 @@ class HeadGazeWindow: UIWindow {
         NotificationCenter.default.addObserver(self, selector: #selector(applicationDidLoseGaze(_:)), name: .applicationDidLoseGaze, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .applicationDidAcquireGaze, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .headTrackingDisabled, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(phraseSaved), name: .phraseSaved, object: nil)
```
Clstroud, line 58: Nice

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -71,50 +75,50 @@ class HeadGazeWindow: UIWindow {
...
-                phraseSavedView.topAnchor.constraint(greaterThanOrEqualTo: topAnchor),
-                phraseSavedView.leftAnchor.constraint(greaterThanOrEqualTo: leftAnchor,
-                                                      constant: horizontalPadding),
-                phraseSavedView.rightAnchor.constraint(lessThanOrEqualTo: rightAnchor,
+            let phraseSavedView = UINib(nibName: "PhraseSavedView", bundle: .main).instantiate(withOwner: nil, options: nil).first as! PhraseSavedView
+             phraseSavedView.alpha = 0
+            phraseSavedView.setAlertText(text: toastLabelText)
+             self.phraseSavedView = phraseSavedView
+             addSubview(phraseSavedView)
+             phraseSavedView.translatesAutoresizingMaskIntoConstraints = false
```
Clstroud: Can you `Control + I` this? Looks like the indentation is a bit off

Clstroud (changes requested):
Just a few more suggestions in the spirit of generalizing that view, but it's looking much better :)

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -93,28 +92,28 @@ class HeadGazeWindow: UIWindow {
...
-                        UIView.animate(withDuration: fadeInOutDuration,
-                                       delay: presentationDuration,
-                                       options: [.beginFromCurrentState, .curveEaseOut],
-                                       animations: { self?.phraseSavedView?.alpha = 0 },
-                                       completion: { dismissalDidFinish in
-                                        guard dismissalDidFinish else { return }
-                                        self?.phraseSavedView?.removeFromSuperview()
-                        })
-        })
+         let fadeInOutDuration: TimeInterval = 0.5
```
stevefosterwta, line 95: Are there an extra spaces here?

thomasshealy: I don't think so.

Vocable/Common/Views/ToastView.swift
```diff
@@ -22,6 +24,12 @@ class PhraseSavedView: BorderedView {
...
     }
+    
+    var text: String = "" {
+        didSet {
+            titleLabel.text = text
+        }
+    }
 
     override func awakeFromNib() {
         super.awakeFromNib()
```
Clstroud, line 35:
```suggestion
        super.awakeFromNib()
        titleLabel.text = text
```
Ensuring that if the xib is loaded lazily, the titleLabel will get some text if it was already assigned

Vocable/Common/Views/ToastView.swift
```diff
@@ -22,6 +24,12 @@ class PhraseSavedView: BorderedView {
             cornerRadius = frame.height / 2
         }
     }
+    
+    var text: String = "" {
+        didSet {
+            titleLabel.text = text
```
Clstroud:
```suggestion
            titleLabel?.text = text
```
Like the `awakeFromNib`, this ensures that we won't crash in the event that the xib gets lazily loaded and the text is set before titleLabel becomes available. It's not likely in the current implementation, but this class doesn't have knowledge of that :)

Clstroud (approved):
Just two more inline suggestions, but LGTM 👍

Vocable/Common/Views/ToastView.swift
thomasshealy: Good call.

PR #161, opened 2020-03-24
Number Pad Category

### LINKED ISSUE #101: Number pad

jsmorgan42:
Added new number pad category that includes a custom layout.

Screen Shot 2020-03-24 at 2 55 42 PM
Screen Shot 2020-03-24 at 2 55 36 PM
Screen Shot 2020-03-24 at 2 55 23 PM

### DIFF
--- a/Vocable/Features/Presets/Models/TextPresets.swift
+++ b/Vocable/Features/Presets/Models/TextPresets.swift
@@ -9,6 +9,7 @@
 import Foundation
 
 enum PresetCategory: CaseIterable {
+    case numPad
     case category1
     case category2
     case category3
@@ -18,6 +19,8 @@ enum PresetCategory: CaseIterable {
     
     var description: String {
         switch self {
+        case .numPad:
+            return "123 | Yes | No"
         case .category1:
             return "General"
         case .category2:
@@ -35,6 +38,13 @@ enum PresetCategory: CaseIterable {
 }
 
 struct TextPresets {
+    
+    static var numPadCategory: [PhraseViewModel] {
+        var numbers = (1...9).map { PhraseViewModel(unpersistedPhrase: "\($0)")}
+        numbers.append(PhraseViewModel(unpersistedPhrase: "0"))
+        let responses = [PhraseViewModel(unpersistedPhrase: "No"), PhraseViewModel(unpersistedPhrase: "Yes")]
+        return numbers + responses
+    }
     
     static var presetsByCategory: [PresetCategory: [String]] = [
         .category1: ["Please",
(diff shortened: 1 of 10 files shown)


Vocable/Features/Presets/Models/TextPresets.swift
```diff
@@ -18,6 +19,8 @@ enum PresetCategory: CaseIterable {
     
     var description: String {
         switch self {
+        case .numPad:
+            return "123 | Yes | No"
```
Clstroud, line 23: Should the yes/no be localized strings?

Vocable/Features/Presets/Models/TextPresets.swift
```diff
@@ -36,6 +39,13 @@ enum PresetCategory: CaseIterable {
 
 struct TextPresets {
     
+    static var numPadCategory: [PhraseViewModel] {
+        var numbers = (1...9).map { PhraseViewModel(unpersistedPhrase: "\($0)")}
+        numbers.append(PhraseViewModel(unpersistedPhrase: "0"))
+        let responses = [PhraseViewModel(unpersistedPhrase: "No"), PhraseViewModel(unpersistedPhrase: "Yes")]
```
Clstroud: And also these (localization)

Vocable/Features/Presets/PresetPageCollectionViewController.swift
```diff
@@ -109,30 +109,36 @@ class PresetPageCollectionViewController: UICollectionViewController {
...
+        override init(section: NSCollectionLayoutSection) {
+            super.init(section: section)
         }
         
         required init?(coder: NSCoder) {
             fatalError("init(coder:) has not been implemented")
         }
         
-        private static func presetsSectionLayout(for traitCollection: UITraitCollection) -> NSCollectionLayoutSection {
+        class func gridDimensions(for traitCollection: UITraitCollection) -> (numberOfRows: Int, itemsPerRow: Int) {
```
Clstroud:
```suggestion
        fileprivate class func gridDimensions(for traitCollection: UITraitCollection) -> (numberOfRows: Int, itemsPerRow: Int) {
```
I think?

Clstroud (approved):
I think those strings are going to need to be localized so we don't break the incoming localizations, but otherwise this looks good

Vocable/Features/Presets/Models/TextPresets.swift
jsmorgan42: Looks like all these descriptions will be moved out from this file & moved to JSON files in the localization PR, so it might make more sense to keep it how it is now.

PR #173, opened 2020-03-28
Language-specific word predictions and displaying only one "My Sayings" category

pg8wood:
Opening this draft to discuss localization strategy. This is an early working copy, and any suggestions for improvements are more than welcome! 

## Additions: 
- Always use the device locale for the `UITextChecker` predictions. This allows us to predict for any language supported by iOS, even if we haven't localized the preset phrases yet
- Adds a lightweight migration of the Core Data store to add a `locale` property to `Category` and `Phrase` data models
  - Expands the `AppDelegate`'s Core Data routines to give the `locale` a default value when migrating the Core Data store. This is required since we cannot hard-code a default value for the locale in the Core Data models
- Uses `locale` to prevent showing multiple "My Sayings" categories
- Ensures the `AVSpeechSynthesizer` speaks using the correct voice--English for untranslated preset phrases and the device locale's default voice for translated phrases and custom phrases from the keyboard and in "My Sayings"

**Supported localization <translated "My Sayings"> category and localized text predictions**
https://imgur.com/a/qLHwSNy

**Unsupported localization "My Sayings" (non-localized) category and localized text predictions**
https://imgur.com/a/r1Ryunc


Vocable/CoreData/NSManagedObject+Helpers.swift
```diff
@@ -24,16 +24,16 @@ extension NSManagedObjectIdentifiable where Self: NSManagedObject {
         } else {
             fetchRequest.predicate = NSPredicate(format: "identifier == \(identifier)")
         }
-        let result = (try? context.fetch(fetchRequest))?.first
-        return result
+        return (try? context.fetch(fetchRequest))?.first
     }
     
-    static func fetchAll(in context: NSManagedObjectContext, sortDescriptors: [NSSortDescriptor]? = nil) -> [Self] {
+    static func fetchAll(in context: NSManagedObjectContext, locale: String? = Locale.current.identifier, sortDescriptors: [NSSortDescriptor]? = nil) -> [Self] {
```
Clstroud: I don't think this makes sense as a change for all `NSManagedObject` since they aren't guaranteed to have a `locale` property. I would suggest maybe having an optional predicate parameter and then having a locale-aware variant of this function as an extension on `Phrase`/`Category` that specifies a locale via the predicate parameter in this function.

Vocable/AppDelegate.swift
```diff
@@ -82,4 +83,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             }
         }
     }
+    
+    private func migrateUserCreatedContent(in context: NSManagedObjectContext) {
+        Category.fetchAll(in: context, locale: nil).forEach { category in
```
Clstroud: This function says it's migrating user-generated content but it's currently doing _everything_. I assume that's because we don't have locale-aware JSON yet?

Vocable/Features/Presets/PresetPageCollectionViewController.swift
```diff
@@ -95,7 +95,12 @@ class PresetPageCollectionViewController: UICollectionViewController {
 
             // Dispatch to get off the main queue for performance
             DispatchQueue.global(qos: .userInitiated).async {
-                AVSpeechSynthesizer.shared.speak(viewModel.utterance)
+                if viewModel.categories.map({ $0.name }).contains(NSLocalizedString("My Sayings", comment: "Category: My Sayings")) {
```
Clstroud: Can we have a hardcoded identifier for My Sayings so we're not relying on these strings? If we change the name of the category this could break

Vocable/AppDelegate.swift
pg8wood: Re: our in-person conversation: I'll remove this and we can assume English as a default if the locale is nil

Vocable/AppDelegate.swift
```diff
@@ -33,10 +33,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if let url = container.persistentStoreCoordinator.persistentStores.first?.url?.absoluteString.removingPercentEncoding {
             print("NSPersistentStore URL: \(url)")
         }
-        let context = container.viewContext
+
+        let context = container.newBackgroundContext()
```
Clstroud: We should chat about this -- right now the UI doesn't have a means of handling the "I'm migrating Core Data, give me a second." Thinking it might be better to block until we're ready to display data or have a strategy for migration limbo.

Vocable/AppDelegate.swift
```diff
@@ -57,39 +57,50 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
     private func createPrescribedEntities(in context: NSManagedObjectContext) {
+        guard let presetJSON = TextPresets.presets else {
+            assertionFailure("No presets found")
+            return
+        }
 
-        // Create entities that are provided implicitly
-        for presetCategory in TextPresets.presetsByCategory {
+        let preferredLocals = Bundle.main.preferredLocalizations
```
Clstroud: Typo: "Locales"

Clstroud: Also not sure how this compares to `Locale.preferredLanguages` but we should double check the docs to make sure we're using the right thing. @pg8wood might have some insight since he's been brushing up on that stuff more recently.

Vocable/AppDelegate.swift
```diff
@@ -57,39 +57,50 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-                category.isUserGenerated = true
+            for localizedName in presetCategory.localizedName {
+                let locale = preferredLocals.first { $0 == localizedName.key } ?? "en"
+                category.name = presetCategory.localizedName[locale]
             }
+        }
+
+        for presetPhrase in presetJSON.phrases {
+            let phrase = Phrase.fetchOrCreate(in: context, matching: presetPhrase.id)
+            phrase.creationDate = Date()
```
Clstroud: I believe this should check the `.isInserted` property to make sure it is, in fact, a newly created object.

Vocable/AppDelegate.swift
```diff
@@ -57,39 +57,50 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        }
 
-        // Create entities that are provided implicitly
-        for presetCategory in TextPresets.presetsByCategory {
+        let preferredLocals = Bundle.main.preferredLocalizations
 
-            let category = Category.fetchOrCreate(in: context, matching: presetCategory.title)
+        for presetCategory in presetJSON.categories {
+            let category = Category.fetchOrCreate(in: context, matching: presetCategory.id)
             category.creationDate = Date()
```
Clstroud: I believe this should check the `.isInserted` property to make sure it is, in fact, a newly created object.

Vocable/AppDelegate.swift
```diff
@@ -57,39 +57,50 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-            category.name = presetCategory.title
+            category.isUserGenerated = false
 
-            if category.name == TextPresets.savedSayingsIdentifier {
-                category.isUserGenerated = true
+            for localizedName in presetCategory.localizedName {
+                let locale = preferredLocals.first { $0 == localizedName.key } ?? "en"
+                category.name = presetCategory.localizedName[locale]
             }
```
Clstroud: Is this is going to have the right effect? I think you should be getting the first `preferredLocales` element that `presetCategory.localizedName.keys` contains, then use that to get your name string. You could possibly even do that as a `guard` before attempting to fetch/create the managed object.

Vocable/AppDelegate.swift
```diff
@@ -57,39 +57,50 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            phrase.isUserGenerated = false
 
-            for preset in presetCategory.presets.reversed() {
-                let phrase = Phrase.fetchOrCreate(in: context, matching: preset)
-                phrase.creationDate = Date()
-                phrase.utterance = preset
-                phrase.addToCategories(category)
+            for localizedUtterance in presetPhrase.localizedUtterance {
+                let locale = preferredLocals.first { $0 == localizedUtterance.key } ?? "en"
+                phrase.utterance = presetPhrase.localizedUtterance[locale]
```
Clstroud: Same question as above re: how this loop is working. Granted, this one has a `break` so it should have fewer side-effects, but it still looks to me like it would be grabbing the first preset `localizedUtterance` and getting the first `preferredLocale` element that matches its key... which would be the same value? I think?

Vocable/AppConfig.swift
```diff
@@ -24,4 +24,6 @@ struct AppConfig {
     static var isHeadTrackingSupported: Bool {
         return ARFaceTrackingConfiguration.isSupported
     }
+    
+    static let preferredLanguageIdentifier = Locale.preferredLanguages.first ?? Locale.current.identifier
```
Clstroud: I don't think this is needed with the updated strategy? The new strategy is meant to allow us to respect the entire array of preferred languages as fallback values. Whether that's 100% the correct thing to do remains to be seen.

Clstroud (changes requested):

Vocable/AppDelegate.swift
stevefosterwta: I don't think that saving changes from a background context to view context is a bottleneck.

stevefosterwta: That being said using the view context isn't that bad.

Vocable/Features/Presets/PresetsPageViewController.swift
```diff
@@ -29,9 +29,9 @@ class PresetsPageViewController: UIPageViewController, UIPageViewControllerDataS
         var chunked = presetViewModels.chunked(into: itemsPerPage)
         var pageViewControllers: [UIViewController] = []
         
-        if ItemSelection.selectedCategory.name == TextPresets.numPadDescription {
+        if ItemSelection.selectedCategory.identifier == TextPresets.numPadIdentifier {
```
jsmorgan42, line 32: 💯

jsmorgan42 (approved):
🥇

Clstroud (approved):

PR #174, opened 2020-03-30
Feature/window refactor

thomasshealy:
Refactored the Head Gaze Window so that it is no longer responsible for displaying notifications (head tracking lost, phrased saved, etc.).

### DIFF
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -50,23 +50,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private func addObservers() {
         NotificationCenter.default.addObserver(self, selector: #selector(applicationDidLoseGaze(_:)), name: .applicationDidLoseGaze, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .applicationDidAcquireGaze, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .headTrackingDisabled, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(headTrackingDisabled(_:)), name: .headTrackingDisabled, object: nil)
     }
     
     @objc private func applicationDidLoseGaze(_ sender: Any?) {
-        handleWarning(shouldDisplay: true)
+        ToastWindow.shared.presentPersistantWarning(with: NSLocalizedString("Please move closer to the device.", comment: "Warning title when head tracking is lost."))
     }
     
     @objc private func applicationDidAcquireGaze(_ sender: Any?) {
-        handleWarning(shouldDisplay: false)
+        ToastWindow.shared.dismissPersistantWarning()
     }
     
     @objc private func headTrackingDisabled(_ sender: Any?) {
-        handleWarning(shouldDisplay: false)
-    }
-    
-    private func handleWarning(shouldDisplay: Bool) {
-        NotificationWindow.shared.handleWarning(shouldDisplay: shouldDisplay)
+        ToastWindow.shared.dismissPersistantWarning()
     }
 
     private func deleteExistingPrescribedEntities(in context: NSManagedObjectContext) {
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -13,7 +13,6 @@ import CoreData
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
-    var notificationWindow: NotificationWindow?
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
 
@@ -22,14 +21,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         preparePersistentStore()
         
         addObservers()
-        notificationWindow = NotificationWindow(frame: UIScreen.main.bounds)
-        notificationWindow?.isHidden = true
         
         application.isIdleTimerDisabled = true
         let window = HeadGazeWindow(frame: UIScreen.main.bounds)
         window.rootViewController = UIStoryboard(name: "Main", bundle: nil).instantiateInitialViewController()
         window.makeKeyAndVisible()
         self.window = window
+        
         return true
     }
 
@@ -68,7 +66,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     private func handleWarning(shouldDisplay: Bool) {
-        notificationWindow?.handleWarning(shouldDisplay: shouldDisplay)
+        NotificationWindow.shared.handleWarning(shouldDisplay: shouldDisplay)
     }
 
     private func deleteExistingPrescribedEntities(in context: NSManagedObjectContext) {
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -13,13 +13,18 @@ import CoreData
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
+    var notificationWindow: NotificationWindow?
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
 
         // Ensure that the persistent store has the current
         // default presets before presenting UI
         preparePersistentStore()
-
+        
+        addObservers()
+        notificationWindow = NotificationWindow(frame: UIScreen.main.bounds)
+        notificationWindow?.isHidden = true
+        
         application.isIdleTimerDisabled = true
         let window = HeadGazeWindow(frame: UIScreen.main.bounds)
         window.rootViewController = UIStoryboard(name: "Main", bundle: nil).instantiateInitialViewController()
@@ -42,6 +47,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         } catch {
             assertionFailure("Core Data save failure: \(error)")
         }
+    }
+    
+    private func addObservers() {
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidLoseGaze(_:)), name: .applicationDidLoseGaze, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .applicationDidAcquireGaze, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .headTrackingDisabled, object: nil)
+    }
+    
+    @objc private func applicationDidLoseGaze(_ sender: Any?) {
+        handleWarning(shouldDisplay: true)
+    }
+    
+    @objc private func applicationDidAcquireGaze(_ sender: Any?) {
+        handleWarning(shouldDisplay: false)
+    }
+    
+    @objc private func headTrackingDisabled(_ sender: Any?) {
+        handleWarning(shouldDisplay: false)
+    }
+    
+    private func handleWarning(shouldDisplay: Bool) {
+        notificationWindow?.handleWarning(shouldDisplay: shouldDisplay)
     }
 
     private func deleteExistingPrescribedEntities(in context: NSManagedObjectContext) {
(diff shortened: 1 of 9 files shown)


Vocable/AppDelegate.swift
```diff
@@ -43,6 +46,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             assertionFailure("Core Data save failure: \(error)")
         }
     }
+    
+    private func addObservers() {
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidLoseGaze(_:)), name: .applicationDidLoseGaze, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .applicationDidAcquireGaze, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidAcquireGaze(_:)), name: .headTrackingDisabled, object: nil)
```
Clstroud: I almost commented that I thought it was sending a message to the wrong selector, but now I see why you're doing this. Can you make an additional function that handles this case to make it more explicit that we want to dismiss the warning in the event that head tracking is disabled? Maybe even adding a comment for clarity?

Vocable/HeadGazeLib/NotificationWindow.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+//  Created by Thomas Shealy on 3/25/20.
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import UIKit
+import ARKit
+import Combine
+
+class NotificationWindow: UIWindow {
```
Clstroud: `NotificationWindow ` -> `ToastWindow` for prefix consistency?

Vocable/HeadGazeLib/NotificationWindow.swift
```diff
@@ -0,0 +1,72 @@
...
+    
+    func handlePhraseSaved(toastLabelText: String) {
+        toastContainer.handlePhraseSaved(toastLabelText: toastLabelText)
+    }
+    
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+
+        let hitView = super.hitTest(point, with: event)
+        
+        if NotificationWindow.passThroughTag == hitView?.tag {
```
Clstroud: What effect does this actually have/why is it needed? In which cases should the entire window _not_ return `nil` and/or have `.userInteractionEnabled = false`?

Vocable/HeadGazeLib/NotificationWindow.swift
```diff
@@ -0,0 +1,72 @@
...
+            let shared = NotificationWindow(frame: UIScreen.main.bounds)
+            shared.backgroundColor = .clear
+            shared.translatesAutoresizingMaskIntoConstraints = false
+            shared.rootViewController = ToastContainerViewController()
+            _shared = shared
+        }
+        return _shared!
+    }
+    
+    var toastContainer: ToastContainerViewController {
```
Clstroud: Suggestion: `toastContainer` -> `toastContainerViewController`, since "container" itself doesn't really imply any sort of conventional type in UIKit. Would probably improve readability for callers.

Vocable/HeadGazeLib/NotificationWindow.swift
```diff
@@ -0,0 +1,72 @@
...
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import UIKit
+import ARKit
+import Combine
+
+class NotificationWindow: UIWindow {
+    
+    private var cancellables = Set<AnyCancellable>()
```
Clstroud: This doesn't appear to be used?

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -59,102 +55,22 @@ class HeadGazeWindow: UIWindow {
...
-                         UIView.animate(withDuration: fadeInOutDuration,
-                                        delay: presentationDuration,
-                                        options: [.beginFromCurrentState, .curveEaseOut],
-                                        animations: { self?.phraseSavedView?.alpha = 0 },
-                                        completion: { dismissalDidFinish in
-                                         guard dismissalDidFinish else { return }
-                                         self?.phraseSavedView?.removeFromSuperview()
-                         })
-         })
+      //  handleWarning(shouldDisplay: false)
```
Clstroud: Is this function still relevant?

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -59,102 +55,22 @@ class HeadGazeWindow: UIWindow {
 
     @objc private func applicationDidLoseGaze(_ sender: Any?) {
         cancelCurrentGazeIfNeeded()
-        handleWarning(shouldDisplay: true)
+       // handleWarning(shouldDisplay: true)
     }
     
     @objc private func applicationDidAcquireGaze(_ sender: Any?) {
-        handleWarning(shouldDisplay: false)
+      //  handleWarning(shouldDisplay: false)
```
Clstroud: Is this function still relevant?

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -59,102 +55,22 @@ class HeadGazeWindow: UIWindow {
 
     @objc private func applicationDidLoseGaze(_ sender: Any?) {
         cancelCurrentGazeIfNeeded()
-        handleWarning(shouldDisplay: true)
+       // handleWarning(shouldDisplay: true)
```
Clstroud: Is this function still relevant?

Vocable/HeadGazeLib/NotificationWindow.swift
```diff
@@ -0,0 +1,72 @@
...
+        
+    }
+    
+    func handleWarning(shouldDisplay: Bool) {
+        toastContainer.handleWarning(shouldDisplay: shouldDisplay)
+    }
+    
+    func handlePhraseSaved(toastLabelText: String) {
+        toastContainer.handlePhraseSaved(toastLabelText: toastLabelText)
+    }
```
Clstroud: Anything operating in Toast-land should have no concept of the content being displayed, only the kind of toast that's being presented. Things like "Phrase saved" or "Warning" are concepts that should be confined to the callers. Instead we should have concepts like we discussed in planning such as "ephemeral" or "persistent" with strings and images.

Vocable/Features/Settings/EditSayings/EditKeyboardViewController.swift
```diff
@@ -21,6 +21,8 @@ class EditKeyboardViewController: UIViewController, UICollectionViewDelegate {
     
     var phraseIdentifier: String?
     
+    private var notificationWindow: NotificationWindow?
```
Clstroud: Is this instance used?

Vocable/Features/Presets/PresetsViewController.swift
```diff
@@ -16,6 +16,7 @@ class PresetsViewController: UICollectionViewController {
     
     private var dataSource: UICollectionViewDiffableDataSource<Section, ItemWrapper>!
     private var disposables = Set<AnyCancellable>()
+    private var notificationWindow: NotificationWindow?
```
Clstroud: Is this instance used?

Vocable/Common/ToastContainerViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+
+import UIKit
+
+class ToastContainerViewController: UIViewController {
+    
+    private weak var phraseSavedView: UIView?
+    private weak var warningView: UIView?
+    
+    private var phraseSavedViewVisible = false
+    private var warningViewVisible = false
```
Clstroud: I think these are redundant? The views themselves get `nil`'d out when they are dismissed

Vocable/Common/ToastContainerViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        NotificationWindow.shared.isHidden = false
```
Clstroud: In general the view controller should not be manipulating its parent's window like this. IMO this should be the job of the generic `ToastController` we discussed during planning.

Vocable/Features/Presets/PresetsViewController.swift
```diff
@@ -370,7 +371,7 @@ class PresetsViewController: UICollectionViewController {
 
                 do {
                     try context.save()
-                    (self.view.window as? HeadGazeWindow)?.handlePhraseSaved(toastLabelText: NSLocalizedString("Saved to My Sayings", comment: "Saved to My Sayings"))
+                    NotificationWindow.shared.handlePhraseSaved(toastLabelText: NSLocalizedString("Saved to My Sayings", comment: "Saved to My Sayings"))
```
Clstroud: Another reason for the `ToastController`: it feels clunky for callers to make any assumptions about how Toasts are being presented. The window itself should be a hidden implementation detail.

Clstroud (changes requested):
This is a solid first pass, but there are a handful of things that could use another cleanup pass and/or a little bit more elbow grease to more closely approximate what we discussed during planning. More than happy to chat about specifics 😄

Vocable/HeadGazeLib/NotificationWindow.swift
thomasshealy: Good call, this is a much better way.

Vocable/Common/ToastContainerViewController.swift
thomasshealy: I need them to signal when their respective animations are done.  If I don't then the window can get set to hidden before a fade out of one of the warnings is complete - which is problematic.  I'm open to suggestions on other ways to do this, the other methods I tried led to some messy force unwraps.

Clstroud: This sounds like it would lend itself to the notion of a completion handler in the toast model object (that would hypothetically be stored on the stack in the persistent case). That way the handler can be called when the animation completes and the hypothetical `ToastController` can decide what to do based on how many things are being still displayed/enqueued.

Vocable/Common/ToastContainerViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+            updateWindowVisibility()
+        }
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        updateWindowVisibility()
+    }
+    
+    func updateWindowVisibility() {
```
Clstroud:
```suggestion
    private func updateWindowVisibility() {
```

Vocable/Common/Views/WarningView.swift
```diff
@@ -15,6 +15,5 @@ class WarningView: UIView {
 
     override func awakeFromNib() {
```
Clstroud: This function isn't doing anything anymore and can be deleted

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -48,113 +45,21 @@ class HeadGazeWindow: UIWindow {
...
-                                        options: [.beginFromCurrentState, .curveEaseOut],
-                                        animations: { self?.phraseSavedView?.alpha = 0 },
-                                        completion: { dismissalDidFinish in
-                                         guard dismissalDidFinish else { return }
-                                         self?.phraseSavedView?.removeFromSuperview()
-                         })
-         })
     }
 
     override func addSubview(_ view: UIView) {
```
Clstroud, line 58: This probably isn't doing anything now that we've removed those excess views

Vocable/Common/ToastContainerViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+        if phraseSavedView == nil && warningView == nil {
+            ToastWindow.shared.isHidden = true
+        } else {
+            ToastWindow.shared.isHidden = false
+        }
```
Clstroud, line 34: Per our pairing yesterday: since we know we aren't happy with this `ToastWindow.shared.isHidden` business and are running with it for the time being, can we add a comment saying that a future revision should move away from manipulating that window from here?

Clstroud (approved):

Vocable/HeadGazeLib/HeadGazeWindow.swift
thomasshealy: We still need this to bring the cursor view to the front.

PR #195, opened 2020-04-03
Disable Warning Toast While Using Touch

thomasshealy:
This fixes the problem where the toast which gets displayed after a user loses head tracking still displays while the user is using touch.

### DIFF
--- a/Vocable.xcodeproj/project.pbxproj
+++ b/Vocable.xcodeproj/project.pbxproj
@@ -1132,8 +1132,8 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
-				CODE_SIGN_IDENTITY = "iPhone Developer";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_IDENTITY = "Apple Development";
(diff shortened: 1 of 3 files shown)


Clstroud:
As promised, here is a crude (and I mean _crude_) pseudocode example for what we were discussing earlier today:

```
private let duration: TimeInterval = 3
private var lastTouchDate: Date?
private var trackingDisabledByTouch = false

private func newTouchHappened() {
    scheduleNewTouchTimerCompletionIfNeeded()
}

private func scheduleNewTouchTimerCompletionIfNeeded() {
    guard trackingDisabledByTouch == false else { return }
    trackingDisabledByTouch = true

    // using an inline function that can be called recursively without
    // worrying about the guard above
    func schedule(duration: TimeInterval = self.duration) {

        DispatchQueue.main.asyncAfter(deadline: .now() + duration) {
            guard let lastTouchDate = self.lastTouchDate else { return }
            let passed = Date().timeIntervalSince(lastTouchDate)
            if passed >= self.duration {
                self.trackingDisabledByTouch = false
                // re-enable tracking, re-present warning if needed
            } else {
                schedule(duration: self.duration - passed)
            }
        }
    }
}
```

I'm sure that's got some rough edges, but it's something to chew on.

PR #199, opened 2020-04-06
Toast Warning Fix

thomasshealy:
This fixes the problem where the toast which gets displayed after a user loses head tracking still displays while the user is using touch.

### DIFF
--- a/Vocable/Common/ToastContainerViewController.swift
+++ b/Vocable/Common/ToastContainerViewController.swift
@@ -9,6 +9,8 @@
 import UIKit
 
 class ToastContainerViewController: UIViewController {
+    var isToastWarningVisible = false
+    
     private weak var phraseSavedView: UIView? {
         didSet {
             updateWindowVisibility()
@@ -99,9 +101,11 @@ class ToastContainerViewController: UIViewController {
         }
         let alphaValue = shouldDisplay ? 1.0 : 0.0
         UIView.animate(withDuration: 0.5, delay: 0, options: .beginFromCurrentState, animations: {
+            self.isToastWarningVisible = true
             self.warningView?.alpha = CGFloat(alphaValue)
         }, completion: { [weak self] didFinish in
             if didFinish && !shouldDisplay {
+                self?.isToastWarningVisible = false
                 self?.warningView?.removeFromSuperview()
             }
         })
--- a/Vocable/Common/ToastContainerViewController.swift
+++ b/Vocable/Common/ToastContainerViewController.swift
@@ -9,7 +9,8 @@
 import UIKit
 
 class ToastContainerViewController: UIViewController {
-    var isToastWarningVisible = false
+    private var isToastWarningVisible = false
+    private var headTrackingLostText: String?
     
     private weak var phraseSavedView: UIView? {
         didSet {
@@ -27,7 +28,6 @@ class ToastContainerViewController: UIViewController {
         updateWindowVisibility()
     }
     
-    //In the future we should get away from manipulating the window here.
     private func updateWindowVisibility() {
         if phraseSavedView == nil && warningView == nil {
             ToastWindow.shared.isHidden = true
@@ -36,6 +36,7 @@ class ToastContainerViewController: UIViewController {
         }
     }
     
+    //In the future we should get away from manipulating the window here.
     func handlePhraseSaved(toastLabelText: String) {
         if phraseSavedView == nil {
             let phraseSavedView = UINib(nibName: "ToastView", bundle: .main).instantiate(withOwner: nil, options: nil).first as! ToastView
@@ -99,14 +100,17 @@ class ToastContainerViewController: UIViewController {
                 warningView.centerXAnchor.constraint(equalTo: view.safeAreaLayoutGuide.centerXAnchor)
             ])
         }
+        if headTrackingLostText != nil && headTrackingLostText == title {
+            return
+        }
+        self.headTrackingLostText = title
         let alphaValue = shouldDisplay ? 1.0 : 0.0
         UIView.animate(withDuration: 0.5, delay: 0, options: .beginFromCurrentState, animations: {
-            self.isToastWarningVisible = true
             self.warningView?.alpha = CGFloat(alphaValue)
         }, completion: { [weak self] didFinish in
             if didFinish && !shouldDisplay {
-                self?.isToastWarningVisible = false
                 self?.warningView?.removeFromSuperview()
+                self?.headTrackingLostText = nil
             }
         })
     }
--- a/Vocable/HeadGazeLib/HeadGazeWindow.swift
+++ b/Vocable/HeadGazeLib/HeadGazeWindow.swift
@@ -21,6 +21,8 @@ class HeadGazeWindow: UIWindow {
     private var touchGazeDisableBeganDate: Date?
     
     private var cancellables = Set<AnyCancellable>()
+    
+    var trackingDisabledByTouch = false
 
     override init(frame: CGRect) {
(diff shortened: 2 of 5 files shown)


Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -151,15 +180,17 @@ class HeadGazeWindow: UIWindow {
...
                 }
             super.sendEvent(originalEvent)
             return
         }
 
         if let gazeDisabledStart = touchGazeDisableBeganDate {
-            if Date().timeIntervalSince(gazeDisabledStart) >= touchGazeDisableDuration {
+            if Date().timeIntervalSince(gazeDisabledStart) >= touchGazeDisableDuration && trackingDisabledByTouch {
+                newTouchHappened()
+            } else if Date().timeIntervalSince(gazeDisabledStart) >= touchGazeDisableDuration && !trackingDisabledByTouch {
```
Clstroud: This logic feels unnecessarily redundant. Can it be consolidated into a single branch for the `Date().timeIntervalSince(gazeDisabledStart) >= touchGazeDisableDuration` and then handle if/else for `trackingDisabledByTouch` inside?

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -62,6 +64,33 @@ class HeadGazeWindow: UIWindow {
             bringSubviewToFront(cursorView)
         }
     }
+    
+    private func newTouchHappened() {
+        scheduleNewTouchTimerCompletionIfNeeded()
+    }
```
Clstroud:
```suggestion
```

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -151,15 +180,17 @@ class HeadGazeWindow: UIWindow {
                 if originalEvent.type == .touches {
                     extendGazeDisabledPeriodForTouchEvent()
                     cursorView?.setCursorViewsHidden(true, animated: true)
+                    newTouchHappened()
```
Clstroud:
```suggestion
                    scheduleNewTouchTimerCompletionIfNeeded()
```

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -62,6 +64,33 @@ class HeadGazeWindow: UIWindow {
...
+        func schedule(duration: TimeInterval = self.touchGazeDisableDuration) {
+            DispatchQueue.main.asyncAfter(deadline: .now() + duration) {
+                guard let lastTouchDate = self.touchGazeDisableBeganDate else { return }
+                let passed = Date().timeIntervalSince(lastTouchDate)
+                if passed >= self.touchGazeDisableDuration {
+                    self.cursorView?.setCursorViewsHidden(false, animated: true)
+                    self.touchGazeDisableBeganDate = nil
+                    self.trackingDisabledByTouch = false
+                    if !UIApplication.shared.isGazeTrackingActive {
+                        ToastWindow.shared.presentPersistantWarning(with: NSLocalizedString("Please move closer to the device.", comment: "Warning title when head tracking is lost."))
```
Clstroud: This presentation happens in multiple places. I'm thinking it would be cleaner to have a `updateHeadTrackingWarningToast()` function that handles presenting/dismissing based on the touch tracking state/gaze disable duration, etc.

Doing so would prevent this function from needing to worry about explicitly presenting/dismissing the toast, as on line 75 (which is not really within the scope of its responsibility anyway).

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -62,6 +64,33 @@ class HeadGazeWindow: UIWindow {
             bringSubviewToFront(cursorView)
         }
     }
+    
+    private func newTouchHappened() {
+        scheduleNewTouchTimerCompletionIfNeeded()
+    }
+    
+    private func scheduleNewTouchTimerCompletionIfNeeded() {
```
Clstroud: Honestly, I think most of this logic belongs in `extendGazeDisabledPeriodForTouchEvent()` as opposed to a new function. The pseudocode was only meant to provide the general logic. As it stands, there is too much overlap in responsibility between the code in `sendEvent`, `extendGazeDisabledPeriodForTouchEvent()`, and this new function. I think it can be consolidated a fair bit.

Clstroud (changes requested):

Vocable/HeadGazeLib/HeadGazeWindow.swift
```diff
@@ -158,8 +179,9 @@ class HeadGazeWindow: UIWindow {
...
         if let gazeDisabledStart = touchGazeDisableBeganDate {
             if Date().timeIntervalSince(gazeDisabledStart) >= touchGazeDisableDuration {
-                cursorView?.setCursorViewsHidden(false, animated: true)
-                touchGazeDisableBeganDate = nil
+                if !trackingDisabledByTouch {
+                    cursorView?.setCursorViewsHidden(false, animated: true)
+                }
             } else {
                 // Waiting for touch timeout to allow events to propagate
                 return
```
Clstroud, line 190: Something to check: I think this would work the same if this entire `if let gazeDisabledStart = touchGazeDisableBeganDate` statement turned into simply 
```
if touchGazeDisableBeganDate != nil { 
    // Waiting for touch timeout to allow events to propagate
    return 
}
````

File: Vocable/HeadGazeLib/HeadGazeWindow.swift [hunk 7]
```diff
@@ -74,6 +76,25 @@ class HeadGazeWindow: UIWindow {
...
+        trackingDisabledByTouch = true
+        ToastWindow.shared.dismissPersistantWarning()
+        func schedule(duration: TimeInterval = self.touchGazeDisableDuration) {
+            DispatchQueue.main.asyncAfter(deadline: .now() + duration) {
+                guard let lastTouchDate = self.touchGazeDisableBeganDate else { return }
+                let passed = Date().timeIntervalSince(lastTouchDate)
+                if passed >= self.touchGazeDisableDuration {
+                    self.cursorView?.setCursorViewsHidden(false, animated: true)
+                    self.touchGazeDisableBeganDate = nil
+                    self.trackingDisabledByTouch = false
```
Clstroud: Given how these are set in tandem, I think they are redundant. `trackingDisabledByTouch` is great for communicating your intent, but I believe it could be a computed property that simply returns `touchGazeDisableBeganDate != nil`

Vocable/HeadGazeLib/HeadGazeWindow.swift
Clstroud: And if the computed property thing pans out, it would read even better as checking that boolean 😉

Vocable/Common/ToastContainerViewController.swift
```diff
@@ -9,6 +9,8 @@
 import UIKit
 
 class ToastContainerViewController: UIViewController {
+    var isToastWarningVisible = false
```
Clstroud: Externalizing this bit of state feels unnecessary. IMO this view controller should track its own state of what's being displayed and make it a no-op if a consumer attempts to re-present something that's already being presented. Originally we had discussed that being some sort of Toast model object, but for now it could simply be the string that it's given to present.

Clstroud (changes requested):
So much cleaner! Just a couple more things that shouldn't be a huge lift, but I think they'll make it even more concise/clearer. Good work @thomasshealy 😄

Clstroud (approved):

PR #218, opened 2020-04-14
Feature/german keyboard

jsmorgan42:
Changes included:
- Refactored keyboard layout to move dynamically change layout given a number of rows & columns
- Change the layout & datasource depending on the active preferred language code

Screen Shot 2020-04-14 at 3 20 28 PM
Screen Shot 2020-04-14 at 3 18 12 PM
Screen Shot 2020-04-14 at 3 19 47 PM


Vocable/Features/Presets/PresetUICollectionViewCompositionalLayout.swift
```diff
@@ -338,67 +338,83 @@ class PresetUICollectionViewCompositionalLayout: UICollectionViewCompositionalLa
...
-        
-        let section = NSCollectionLayoutSection(group: overallContainerGroup)
+    static func keyboardLayout(with environment: NSCollectionLayoutEnvironment) -> NSCollectionLayoutSection {
+        let traitCollection = environment.traitCollection
+        let dimensions: (rows: Int, columns: Int)
+        
+        let code = KeyboardLocale.current.languageCode
+        switch code {
+        case .en:
+            if traitCollection.horizontalSizeClass == .compact && traitCollection.verticalSizeClass == .regular {
```
Clstroud: Can the evaluated statement of this `if` be captured as a bool up above so it doesn't need to repeat and better communicates what it's checking for? I think it would help with the last tricky part of these changes in terms of readability

Vocable/Features/Presets/PresetsViewController.swift
```diff
@@ -254,9 +251,9 @@ class PresetsViewController: UICollectionViewController {
             
             snapshot.appendSections([.keyboard])
             if traitCollection.horizontalSizeClass == .compact && traitCollection.verticalSizeClass == .regular {
-                snapshot.appendItems(KeyboardKeys.alphabetical.map { ItemWrapper.key("\($0)") })
+                snapshot.appendItems(KeyboardLocale.current.compactPortrait.map { ItemWrapper.key("\($0)") })
             } else {
-                snapshot.appendItems(KeyboardKeys.qwerty.map { ItemWrapper.key("\($0)") })
+                snapshot.appendItems(KeyboardLocale.current.landscape.map { ItemWrapper.key("\($0)") })
```
Clstroud: In this context, `.landscape` and `.compactPortrait` are not obviously returning an array of anything. Can we name them any more clearly?

Also I'm wondering if something like `compactWidth`/`regularWidth` would capture their use cases better than `landscape` and `compactPortrait`? Not sure, just thinking aloud.

Clstroud (approved):
Suggested a couple of tweaks, but this cleaned up nicely 💯 🏅

Vocable/Features/Presets/PresetsViewController.swift
jsmorgan42: The issue with naming them compactWidth & regularWidth is that there are situations when smaller iphones are in landscape & it's still compact width.

PR #222, opened 2020-04-18
User favorites empty states

Clstroud:
Started this work because it felt strange not having it while debugging other things.

It's largely finished; just needs a little massaging and inclusion on some other screens (like the main presets screen).


stevefosterwta (approved):
👍

PR #225, opened 2020-04-20
Bugfix/ipad edit category crash

jsmorgan42:
Changes included
- Added multiple guard statements to protect any divide by zero crashes within the layout
- Fixed constraint on the collection view in edit categories storyboard so that collection view did not have a height of 0 on iPad


stevefosterwta (approved):
🥇 👍

PR #227, opened 2020-04-20
Update Phrases predicate for more reliable updates

Clstroud:
* Uses the NSManagedObject itself in the comparison predicate rather than its objectID


stevefosterwta (approved):
🗡️ 👍

PR #236, opened 2020-04-21
Modifying suggestive text iPhone layout to match designs

### LINKED ISSUE #235: Suggestive text should be separated on iPhone
The suggestive text options should be more separated with no decoration background for iPhone landscape. Correct design shown below:

Screen Shot 2020-04-21 at 1 40 20 PM

jsmorgan42:
The suggestive text options for landscape on iPhone did not match designs, the buttons were a lot smaller making it more difficult to select an option.

Before:
IMG_0813

After:
IMG_0812


Clstroud (approved):
Looks way better 💯

PR #237, opened 2020-04-21
Cancel gaze target on ordinal buttons when reordering category cells

Clstroud:
Fixes an issue where the category ordinal buttons could reorder indefinitely when using gaze


stevefosterwta (approved):

stevefosterwta (approved):
🏁 👍

jsmorgan42 (approved):

PR #243, opened 2020-04-23
Adding UI automation to the Vocable project

stechkev:
Adding UI automation to the Vocable project and corresponding Circle


Clstroud (approved):
💯

PR #262, opened 2020-04-24
Fix main screen layout twitch during launch

Clstroud:
launch

The GIF speaks for itself.

I'm not sure why this was perplexing to track down in the past, but I stumbled into the solution while working on something else. The refactor will obviate all of this anyway, but I figured I should PR it just in case.


stevefosterwta (approved):
👍

Clstroud:
For posterity, here's a GIF of the old erratic behavior for comparison twitchy

stechkev:
Test approved! 👍

PR #264, opened 2020-04-24
Relocate head tracking UI to independent windows

Clstroud:
This has been needed for quite some time, but it will be necessary for the upcoming cleanup work.

Rather than house everything in a single window, this change relocates both the head tracking SceneKit UI and the virtual cursor into their own respective UIWindow instances. This will allow the UI to operate with less top-level view controller containment and (finally) allow modal presentations to remove the presenting view controller's view from the view hierarchy. The latter has been quite taxing on older iPad/iPhone models.


stevefosterwta (approved):
🎧 👍

stechkev:
Test approved.

PR #272, opened 2020-04-27
Main screen test setup

stechkev:
quick and dirty setup of basic tests for the main screen of the app. Also added a new test helper class to hold the default setup and teardown methods, as well as screenshot functionality for failures.

Things to know are that there is a big refactor coming so a lot of the references will be overwritten or changed. As part of that we are going to introduce a bunch of accessibility ids so that we don't have to have crazy view heirarchy references all over the place.


VocableUITests/Screens/MainScreen.swift
```diff
@@ -0,0 +1,23 @@
...
+//  Created by Kevin Stechler on 4/27/20.
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import XCTest
+
+class MainScreen: TestHelper {
+    let defaultCategories = ["General", "Basic Needs", "Personal Care", "Conversation", "Environment", "123", "My Sayings"]
+    let defaultPhraseGeneral = ["Please be patient", "I don't know", "Maybe", "Yes", "I didn't mean to say that", "Please wait", "No", "Thank you"]
+    let defaultPhraseBasicNeeds = ["I want to sit up", "I am finished", "I am uncomfortable", "I am fine", "I want to lie down", "I am in pain", "I am good", "I am tired"]
```
Clstroud, line 14: Is there a way we can leverage identifiers for categories rather than strings? Moreover: should we?

Any changes to the presets content will break UI tests.

stechkev: Ideally, yes we should use identifiers in several places in this. right now, considering the refactor, i wanted to get things up and running. since a lot of things will break with the refactor anyway, I think that's a good time to add in identifiers and fix the tests.

VocableUITests/Tests/TestHelper.swift
```diff
@@ -0,0 +1,31 @@
...
+//  VocableUITests
+//
+//  Created by Kevin Stechler on 4/27/20.
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import Foundation
+import XCTest
+
+class TestHelper: XCTestCase {
```
nerdsosky: Should go ahead and rename this to BaseTest so it's clear what this is going to be based on our discussion yesterday

nerdsosky (approved):
Nice, this is a good starting point. Just one minor thing to fix

VocableUITests/Tests/MainScreenTests.swift
```diff
@@ -0,0 +1,38 @@
...
+//  MainScreenTests.swift
+//  VocableUITests
+//
+//  Created by Kevin Stechler on 4/27/20.
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import XCTest
+
+class MainScreenTests: BaseScreen {
```
nerdsosky: Test classes should have their own BaseTest instead of BaseScreen

VocableUITests/Tests/BaseScreen.swift
```diff
@@ -0,0 +1,31 @@
...
+//  VocableUITests
+//
+//  Created by Kevin Stechler on 4/27/20.
+//  Copyright © 2020 WillowTree. All rights reserved.
+//
+
+import Foundation
+import XCTest
+
+class BaseScreen: XCTestCase {
```
nerdsosky: BaseScreen doesn't need to inherit from anything it can be it's own class. The setup/teardown functions should go in the BaseTest class along with the captureFailure. Eventually BaseScreen will house any common functionality across multiple screens.

nerdsosky (changes requested):
Added some more comments

nerdsosky (approved):

PR #284, opened 2020-04-29
Feature flagged Add Category

stevefosterwta:
- Adds the edit category button to the environment variable `EditPhrasesEnabled`
- Also fixes the relationship between categories and phrases. All user created phrases would display in both My Sayings and any custom categories.


Vocable/Utilities/Phrase+Helpers.swift
```diff
@@ -25,4 +26,17 @@ extension Phrase {
...
+        phrase.isUserGenerated = true
+        phrase.creationDate = Date()
+        phrase.lastSpokenDate = Date()
+        phrase.utterance = text
+        phrase.languageCode = AppConfig.activePreferredLanguageCode
+        phrase.addToCategories(category)
+        category.addToPhrases(phrase)
+        return phrase
+    }
+
```
stevefosterwta, line 41: Maybe `Category` _should_ be an optional parameter in the original function.

Clstroud: At the very least, the original version should call into this one after it has fetched the user favorites category

Clstroud (approved):
Just the one comment, but otherwise LGTM

PR #286, opened 2020-04-30
Introduce Italian keyboard layout

Clstroud:
Note: This will only affect the modal text editor at the current stage of the refactor. 

The portrait layout on iPad is not the best, but it’s consistent with the existing layouts for English/German so I think any changes there would be a separate effort.


stevefosterwta (approved):
🇮🇹 👍

PR #310, opened 2020-05-06
Change Category -> Phrase relationship to be one-to-one

Clstroud:
# Description of Work
* Fixes an issue where deleting a custom category would crash if the deletion results in orphaned phrases
  - Makes the relationship one-to-one
  - Ensures category deletions cascade down to their phrases


stevefosterwta (approved):
👨‍💻 👍

PR #318, opened 2020-05-07
Fix infinite scroll not starting in center

Clstroud:
# Description of Work
* Fixes an issue where the last-invalidated contentSize was being cached despite the scroll view not actually being scrolled to its initial state. With this change, the value is not cached or marked as invalidated until it successfully scrolls to that initial state.


stevefosterwta (approved):
🐌 👍

stechkev:
Issue:
when you first load the app, immediately pressing the left navigation arrow doesn't scroll. Pressing the right navigation arrow doesn't scroll on the first tap

stechkev:
test approved

PR #319, opened 2020-05-07
Gazeable Alert Enhancements

Clstroud:
# Description of Work
* Prevents alert buttons from shrinking when touched
* Prevents alert buttons from rendering a background color in their corners (must be transparent)
* Adds a new alert transition animation to more closely match the system animation, per last design review
* Fixes an issue where the external links were not performing the correct action when selected
 
Transition looks like this:
alert


stechkev:
test approved

stevefosterwta (approved):
⛹️‍♂️ 🤖 👍

PR #340, opened 2020-05-08
Localization auto-export

Clstroud:
After running the normal test suite, this PR makes it so that localized strings are automatically extracted into the `CrowdinExport/en.xliff` file so Crowdin can pick it up next time it runs its periodic sync. If there is no diff between the xliff files, no commits are appended to the PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,10 +62,9 @@ jobs:
             git config credential.helper 'cache --timeout=120'
             git config user.email "localization_bot@willowtreeapps.com"
             git config user.name "Localization Bot"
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -46,7 +46,26 @@ jobs:
           command: bundle exec fastlane $FASTLANE_LANE
       - store_artifacts:
           path: output
-          
+  export-localizations-for-crowdin:
+    macos: 
+      xcode: 11.4.0
```
stevefosterwta, line 51: Does this have to be hardcoded?

stevefosterwta (approved):
Awesome. 👍

PR #342, opened 2020-05-11
Removed Carthage

stevefosterwta:
This removes Carthage. Carthage was used to manage the Vocable-Preset dependency.


BarryBryant (approved):
💰 💰 💰 💰

PR #348, opened 2020-05-11
CircleCI: Skip committing if no changes exist to resolve git exit code 1

Clstroud:
We received an incorrect failure status after merging into develop. No commit should be created if there are no changes, which will prevent the error from occurring. This was already present in the other commit script but I missed it in the other.


stevefosterwta (approved):
🔬 👍

PR #354, opened 2020-05-13
Enables feature flags for custom categories/phrases and empty states

stevefosterwta:
This enables features flags for custom categories/phrases and empty states for beta testing.


Clstroud (approved):
I’ll approve but I feel like we should remove the flags if we’re intending to actually enable that feature

PR #364, opened 2020-05-21
A fix for broken head tracking

stevefosterwta:
A fix for broken head tracking.


Clstroud (approved):
This was a great idea you had

PR #366, opened 2020-05-22
Fix root screen disappearing during transition

### LINKED ISSUE #365: Root screen disappearing during transition back from Settings flow
**Describe the bug**
See GIF

**To Reproduce**
- Go into settings flow
- Exit settings flow

**Expected behavior**
The root screen to be visible during the transition back from the settings flow

**Actual behavior**
The root screen is not visible during the transition back from the settings flow

**Screenshots**
image

**Device Information**
 - Device: iPhone 11 Pro
 - OS: iOS 13.4.1

Clstroud:
closes #365 

# Description 

There were a plethora of constraint issues during the transition that seemed to stem from the window being swapped after the rootViewController was installed. This strategy ensures that the window is always the correct type from the start. 

dismissal

### DIFF
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -14,30 +14,30 @@ import AVFoundation
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
-    var window: UIWindow?
+    var window: UIWindow? {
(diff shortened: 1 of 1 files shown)


stevefosterwta (approved):
This was a great idea you had. 👍

PR #378, opened 2020-06-05
Update Crowdin import to pull develop before integrating translations

Clstroud:
Attempts to pull develop into the branch manually to avoid the situation where Crowdin makes a PR based on an older commit than develop's head. This will ensure that the latest localized keys are present so xcodebuild can integrate the translations.


stevefosterwta (approved):
👍

Issue #336, opened 2020-05-08
Edit Phrases in a Custom Category


rc-young:
**As a user, I need the ability to edit existing custom phrases to maximize the customization of my experience.**

**AC**
1. User will have the ability to view existing phrases within the custom category and enter an edit flow to update the phrase
2. Leaving the flow without changes should allow the user to exit without a warning
3. Attempting to exit without saving changes should prompt the user to save those changes. 

Designs:
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=1%3A1344

SashankPatel:
Implemented on PR391

Issue #335, opened 2020-05-08
Add Phrases to Custom Category


rc-young:
**As a user, I would like the ability to create new phrases inside of my custom category so that I may use them from the main screen of Vocable**

Acceptance criteria:

1. To access custom phrases, a user will navigate to the "Categories and Phrases" screen in Settings. Custom categories will have an additional detail screen (accessed by the right arrow) where they will have an option to add new phrases.
2. Custom phrases are unique entities and identical phrases can exist in multiple categories. Deleting or editing an identical phrase from one category does not remove it or edit it in another category.
3. Clicking the "add" button will begin the add/edit keyboard flow. If users exit the keyboard without saving, a discard progress alert will be triggered.
4. User can add phrases to the custom category from the custom category flow, or from the keyboard "add" button.

Implementation Note: Successful save notification will need to be updated. Currently if a phrase is added to a custom category notification says "Saved to My Sayings" but it was actually saved to the custom category. Note when a new category is added the notification says "Changes saved", we may want to use this generic message instead.

Test Note:
Custom Phrases will be in English only as of this ticket.

Designs:
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=1%3A1344

SashankPatel:
Implemented on PR391

Issue #334, opened 2020-05-08
Delete phrases from a custom category


rc-young:
**As a user, I need the ability to delete a custom phrase from a custom category so that I can manage what's available.**

**AC:**
1. User can delete all custom phrases from a custom category but only individually.
2. Users should be presented with a confirmation modal before deletion is completed. 
3. Deleted phrases cannot be recovered for now

**Designs:**
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=1%3A1344

SashankPatel:
Implemented on PR391

Issue #153, opened 2020-03-20
Update toast notification to match designs


Clstroud:
* Add blur/visual effect view to the background
* Add an icon to reinforce the message
<img width="407" alt="Screen Shot 2020-03-20 at 9 46 27 AM" src="https://user-images.githubusercontent.com/143916/77169340-b4898380-6a8f-11ea-8866-729c018d9780.png">
<img width="198" alt="Screen Shot 2020-03-20 at 9 46 33 AM" src="https://user-images.githubusercontent.com/143916/77169343-b5221a00-6a8f-11ea-82a4-831d098288f8.png">

jennydoesthings:
This toast is no longer relevant.

PR #391, opened 2020-08-21
Vocable 362: Add/Edit/Delete Custom Category Phrases

stevefosterwta:
Add custom categories: https://github.com/willowtreeapps/vocable-ios/issues/362
- To add a new category, the user will tap the "plus" icon in the top right-hand corner of the Categories screen. This will take them into the add keyboard flow. New categories will be added to the bottom of the categories list, and should appear on the home screen.
- If a user closes the keyboard without saving, they should receive the discard edits alert.

Add phrase to custom category: https://github.com/willowtreeapps/vocable-ios/issues/335
- To access custom phrases, a user will navigate to the "Categories and Phrases" screen in Settings. Custom categories will have an additional detail screen (accessed by the right arrow) where they will have an option to add new phrases.
- Custom phrases are unique entities and identical phrases can exist in multiple categories. Deleting or editing an identical phrase from one category does not remove it or edit it in another category.
- Clicking the "add" button will begin the add/edit keyboard flow. If users exit the keyboard without saving, a discard progress alert will be triggered.
- User can add phrases to the custom category from the custom category flow, or from the keyboard "add" button.
 - Implementation Note: Successful save notification will need to be updated. Currently if a phrase is added to a custom category notification says "Saved to My Sayings" but it was actually saved to the custom category. Note when a new category is added the notification says "Changes saved", we may want to use this generic message instead.

Edit phrase in custom category: https://github.com/willowtreeapps/vocable-ios/issues/336
- User will have the ability to view existing phrases within the custom category and enter an edit flow to update the phrase
- Leaving the flow without changes should allow the user to exit without a warning
- Attempting to exit without saving changes should prompt the user to save those changes.

Delete phrase from custom category: https://github.com/willowtreeapps/vocable-ios/issues/334
- User can delete all custom phrases from a custom category but only individually.
- Users should be presented with a confirmation modal before deletion is completed.
- Deleted phrases cannot be recovered for now


SashankPatel:
Test Notes: 8/21
- Verified the following in English.
#362 :
New category can be added.  It gets added to the end of the list and if closing keyboard while still editing notification appears.
Issue 1: Save notification is different than designs - there is an enhancement ticket for this #153 
- Tested Rotations no issues.

#335:
- Verified custom phrases can be added/editd.
- Verified the same phrase can be in different categories and they are updated/deleted independently.
- Verified close confirmation.

#336:
- Verified edit with discard notification.

#334:
- Verified Individual delete and delete notification.
- Verified Custom Phrases appear in English only.

Vocable/Supporting Files/en.lproj/Localizable.strings
```diff
@@ -5,7 +5,7 @@
 "category_editor.alert.delete_phrase_confirmation.button.delete.title" = "Delete";
 
 /* Delete phrase confirmation alert title */
-"category_editor.alert.delete_phrase_confirmation.title" = "Deleted phrases cannot be recovered.";
+"category_editor.alert.delete_phrase_confirmation.title" = "Are you sure you want to delete? Removed phrases cannot be restored.";
```
Clstroud, line 8: Only comment I have on all of this: I have a feeling it might be good to avoid mixing delete/removed here to keep all the verbiage consistent?

Clstroud (approved):

PR #401, opened 2020-09-15
Adding some keyboard tests

SashankPatel:
# Description of Work
Added some accessibility identifiers in order to add some keyboard tests to be in parity with iOS.  This was a first pass at the work, it can be improved on.

When adding the accessibility identifier, it looks like not only was it added to the image buttons but also the Y, Z and "," buttons, not sure why though.

TODO: it would be nice if the accessibility was only on the image buttons.
---


SashankPatel:
Closing out this PR, PR479 fixes the keyboard screen tests.

PR #404, opened 2020-09-17
Only enable Firebase Crashlytics for Release builds

nerdsosky:
# Description of Work
We only want crash reports for release builds so this disables them for the Development scheme in Xcode. Builds made with the AdHoc scheme or AppStore scheme will still report crashes

# Notes to Test (Optional)
Run a build using the Development scheme locally on simulator and view the console logs, you should not see any messages about Firebase/Analytics starting up. You can then run another build via the AdHoc/AppStore scheme and view the initialization of Firebase in the console logs


nerdsosky:
Added github to known hosts based on support article from CircleCI here https://support.circleci.com/hc/en-us/articles/360044709573

PR #406, opened 2020-10-12
Add Recents category

### LINKED ISSUE #359: Add a "Recent" Category to the Home Screen
**As a user, I would like to easily access the phrases I use most frequently all in one category on the home screen.**

**Acceptance Criteria**:

- Add a "Recent" category to the Home Screen (with clock icon).
- The 9 most recently used phrases (excluding the number pad) should automatically appear in this category, without the user needing to add them.
- The "Recent" category can reordered and hidden in the category list, but a user cannot tap the arrow to see the detail view.
- An error state should be shown if there are no frequently used phrases.

**Design**:

https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=1%3A141

SpudLaw:
https://github.com/willowtreeapps/vocable-ios/issues/359

closes #359

# Description of Work
Added recents category, also added localized strings

## Notes to Test (Optional)
Naming things is hard, lemme know if you have any suggestions (is bodyLabel okay?? idk????) Also I wanna talk about refactoring the EmptyStateView so that we check for if things are nil before putting it in the stack view ty!!!!!

---


Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -47,34 +47,51 @@ class EmptyStateView: UIView {
...
 
-    private lazy var stackView = UIStackView(arrangedSubviews: [self.imageView, self.label])
+    private lazy var stackView = UIStackView(arrangedSubviews: [self.imageView, self.headerLabel, self.bodyLabel])
 
     init(text: String, image: UIImage? = nil, action: ButtonConfiguration? = nil) {
         self.action = action
         super.init(frame: .zero)
         self.text = text
         self.image = image
+        self.bodyAttributedText = nil
```
SpudLaw: this works, but it's not great pls help meee

Vocable/Features/Root/CategoryDetailViewController.swift
```diff
@@ -28,8 +28,14 @@ class CategoryDetailViewController: PagingCarouselViewController, NSFetchedResul
 
     private lazy var fetchRequest: NSFetchRequest<Phrase> = {
         let request: NSFetchRequest<Phrase> = Phrase.fetchRequest()
-        request.predicate = NSComparisonPredicate(\Phrase.category, .equalTo, self.category)
-        request.sortDescriptors = [NSSortDescriptor(keyPath: \Phrase.creationDate, ascending: false)]
+
+        if category.identifier == Category.Identifier.recents {
+            request.sortDescriptors = [NSSortDescriptor(keyPath: \Phrase.lastSpokenDate, ascending: false)]
+            request.fetchLimit = 9
```
SpudLaw, line 34: also we should talk about this and how the fetchLimit should probably be something else :x

Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -47,34 +47,51 @@ class EmptyStateView: UIView {
...
+            bodyLabel.attributedText
+        }
+        set {
+            bodyLabel.attributedText = newValue
+            if newValue != nil {
+                stackView.insertSubview(bodyLabel, belowSubview: headerLabel)
+            } else {
+                stackView.removeArrangedSubview(bodyLabel)
+                bodyLabel.removeFromSuperview()
+            }
```
Clstroud: If this and the other snippet that does the same for the image were moved into a function, you could call that function from `commonInit()` and from their respective `didSet`s, which would avoid the need to trigger this side effect by redundantly setting those properties to `nil` in the initializers.

Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -47,34 +47,51 @@ class EmptyStateView: UIView {
...
-            label.attributedText
+            headerLabel.attributedText
         }
         set {
-            label.attributedText = newValue
+            headerLabel.attributedText = newValue
+        }
+    }
+
+    var bodyAttributedText: NSAttributedString? {
```
Clstroud: If this is named `bodyAttributedText` it might make sense to rename `attributedText` to `headerAttributedText` to maintain that level of specificity and clarity. Same for the other properties, actually.

Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -84,7 +101,17 @@ class EmptyStateView: UIView {
         self.text = nil
         self.image = nil
         self.attributedText = attributedText
+        self.bodyAttributedText = nil
+        commonInit()
+    }
+
+    init(attributedText: NSAttributedString, image: UIImage? = nil, extraAttributedText: NSAttributedString? = nil, action: ButtonConfiguration? = nil) {
```
Clstroud: `attributedText`/`extraAttributedText` naming should stay consistent with the header/body concepts you've introduced elsewhere here

Vocable/Features/Root/CategoryDetailViewController.swift
Clstroud: I'm not sure we're going to find a better solution than just picking a number and sticking to it. If the device rotates and the number of items on screen changes, we wouldn't want to artificially take items away from the user that were _just_ visible before the rotation.

Vocable/Features/Root/CategoryDetailViewController.swift
```diff
@@ -109,12 +115,19 @@ class CategoryDetailViewController: PagingCarouselViewController, NSFetchedResul
...
             lastUtterance = nil
             return
         }
         lastUtterance = utterance
 
+        if category.identifier != Category.Identifier.recents {
+            phrase.lastSpokenDate = Date()
+        }
+
+        try! frc.managedObjectContext.save()
```
Clstroud: This save should only fire if you assign a new `lastSpokenDate`. You should also make it a `try?` just in case.

Vocable/Supporting Files/en.lproj/Localizable.strings
```diff
@@ -135,3 +135,11 @@
 "timing_and_sensitivity.button.high.title" = "High";
 
 "selection_mode.header.title" = "Selection Mode";
+
+"recents_empty_state.header.title" = "No recently used phrases";
+
+"recents_empty_state.body.title" = "Start using your app to see your most recently used phrases here.";
```
Clstroud: Vocable referring to itself as "your app" does not feel right one bit

Clstroud (changes requested):
A handful of small requests, but nothing huge. Overall this is looking great, @CarolineLaw  😄

Vocable/Common/Views/EmptyStateView.swift
SpudLaw: I talked to Ameir and he said we should use a designated initializer here, I think I might go for that. Sound good?

Clstroud: I would be on board with that if those properties weren't externally assignable? I'm not _personally_ a huge fan of requiring everything to be crammed into an initializer, but that's not a hill I'm planning to die on today

SpudLaw: @ameirwillowtree 👀

ameirwillowtree: I would definitely advocate for a change to the size/number of initializers right now and try to simplify what "this thing can be". Perhaps an enum with associated values would be a nice way to require values for the distinct ways this thing can be created.

ameirwillowtree: A big factor is going to be how it looks/works at the call site which I'm not familiar with. So maybe look at those and see what feels right.

Vocable/Common/Views/ToastView.xib
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="16096" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="16097" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
```
pg8wood: We can probably remove changes that just update the toolsVersion

Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -35,58 +35,73 @@ class EmptyStateView: UIView {
...
     private let button = EmptyStateButton(frame: .zero)
     private let action: ButtonConfiguration?
 
-    private lazy var stackView = UIStackView(arrangedSubviews: [self.imageView, self.label])
+    private lazy var stackView = UIStackView(arrangedSubviews: [self.imageView, self.headerLabel, self.bodyLabel])
 
-    init(text: String, image: UIImage? = nil, action: ButtonConfiguration? = nil) {
+    init(headerText: String? =  nil, headerAttributedText: NSAttributedString? = nil,
+         bodyText: String? = nil, bodyAttributedText: NSAttributedString? = nil,
+         image: UIImage? = nil, action: ButtonConfiguration? = nil) {
```
pg8wood: +1 for enumerating the types of `EmptyStateView` we can represent and simplifying this initializer. By doing that we can remove the subclasses and some of the technically impossible states this initializer can represent :+1:

VocableUITests/Tests/CustomCategoriesTests.swift
```diff
@@ -8,17 +8,13 @@
...
     let customCategory = "Createnewcategory"
 
-
     func testAddNewPhrase() {
         let customPhrase = "ddingcustomcategoryphrasetest"
         let confirmationAlert = "Are you sure? Going back before saving will clear any edits made."
-        let createdCustomCategory = ("8. "+customCategory)
-
-        
+        let createdCustomCategory = ("9. "+customCategory)
```
pg8wood, line 17: Talked in person about fixing this in the future 💯

pg8wood (commented):
Getting close! Looks super solid and I think will be ready with the `EmptyStateView` changes. Nice work! 💯

Vocable/Features/Root/CategoryDetailViewController.swift
```diff
@@ -138,9 +138,10 @@ class CategoryDetailViewController: PagingCarouselViewController, NSFetchedResul
         guard collectionView.backgroundView == nil else { return }
         paginationView.isHidden = true
         if category.identifier == Category.Identifier.recents {
-            collectionView.backgroundView = RecentsEmptyStateView()
+            collectionView.backgroundView = EmptyStateView(type: .recents)
         } else {
-            collectionView.backgroundView = PhraseCollectionEmptyStateView(action: addNewPhraseButtonSelected)
+            let buttonTitle = NSLocalizedString("empty_state.button.title", comment: "Empty state Add Phrase button title")
```
SpudLaw: i feel like we could make this better

Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -39,66 +74,46 @@ class EmptyStateView: UIView {
...
-        get {
-            headerLabel.attributedText
-        }
-        set {
-            headerLabel.attributedText = newValue
-        }
-    }
-
-    var bodyText: String? {
+    var titleAttributedText: NSAttributedString? {
```
pg8wood, line 77: We probably no longer need to expose these if they're not being used outside this type, otherwise someone could come in and change our `titleLabel` or `descriptionLabel` text out from under us. Unless I missed something and this is intentional!

Not a big deal though and no reason to hold up this PR any further

Clstroud (approved):
LGTM 🎉 💯

pg8wood (approved):
dancingPotatoes

PR #412, opened 2020-11-03
WIP: Original Category Flow

stevefosterwta:
WIP on bringing back the original category flow.


SpudLaw:
I think we were supposed to fix the size of a button (I think it was the edit category name button), we can do that later, but just gotta remember to do that.

PR #427, opened 2021-01-07
Initial voice interaction experiment

Clstroud:
Adds a (feature flagged) voice interaction prototype. This will be refined in the develop branch so we can easily push out TestFlight builds for stakeholders to evaluate. Builds shipped in the interim will need to have the flag disabled unless we get more granular build settings for TestFlight vs App Store.

Functionally, this adds:
* "Hey Vocable" hot word support
* Continuous transcription while in the "Voice" category tab
* ML-derived responses to the transcribed input

This will need:
* A proper ML model and heuristic from the data science team (forthcoming)
* Design refinement
* Better permissions/error handling
* Ability to configure the feature in Settings
* Likely some code refactoring as this is currently meant more for UX evaluation


Vocable/Extensions/Category+Helpers.swift
```diff
@@ -18,6 +18,7 @@ extension Category {
         case userFavorites = "preset_user_favorites"
         case numPad = "preset_user_keypad"
         case recents = "preset_user_recents"
+        case voice = "Voice"
```
stevefosterwta, line 21: This will need to be localized at some point.

stevefosterwta (approved):
🗣️ 👍

PR #430, opened 2021-01-11
Voice prototype enhancements

Clstroud:
* Match transcribed text color to designs
* Ensure audio session has the proper category before attempting to tap off the input node 
    - This fixes transcription not initializing on physical devices
* Define transcription locale as `en_US` for greater accuracy
* Prevent synthesized speech from being transcribed when the user chooses a response


stevefosterwta (approved):
🙊 👍

PR #431, opened 2021-01-20
Listening mode improvements + settings UI

Clstroud:
#### Includes
* Reverts to a single transcription stream that transitions states, rather than alternating between competing instances
* Switches to AVAudioPlayer rather than AudioServices for listening mode start/stop sound effects
* Adds user-facing preferences to toggle both the listening feature itself and the hot word shortcut
* Introduces an experimental animation for the responses shown in listening mode

#### Still to do
* Microphone/speech recognition authentication state handling
* Transcription UI updates
* ML model pipeline replacement


stevefosterwta (approved):
🙈 🙉 🙊 👍

PR #432, opened 2021-01-21
Speech and microphone permissions

Clstroud:
* Includes stub copy/UI to handle speech and microphone permissions
* Renames "Voice" category to "Listen" and updates identifiers to match
* Treats "Listen" category as a special case in category settings such that showing/hiding it will enable/disable listening mode


stevefosterwta (approved):
🗣️  👍

PR #436, opened 2021-02-15
Listening Mode refinements

Clstroud:
* iOS 14 restriction, due to new ML model requirements
* English-only restriction for listening mode, since we've only trained on English
* Handle speech framework availability status for when the transcription service is unavailable
* Improve empty state management in response VC
* Improved transcription timer cancellation to help prevent an old timer from interfering with a new transcription before the new transcription has had a chance to extend the timer for its own transcription


stevefosterwta (approved):
🟡 🔈 👍

PR #439, opened 2021-03-17
Listening mode transition + grid layout updates 

Clstroud:
* Adds support for the "vertical timeline" style transition we've discussed for listening mode transitions
* Experimentally insets the listening mode content to better establish content hierarchy (I personally think we should do this everywhere)
* Adds support for vertical alignment and relative sizing for the grid layout. This is really only useful for listening mode, but it was easy enough to borrow concepts from composition layout to add support for it.
* Adds support for footer "explainer" text in the settings UI (likely needs a refinement pass) 
* Bug fixes that I encountered along the way (will note them in the comments)

I wanted to clean this up as much as I reasonably could and merge so we can integrate a new version of the VocableListenCore framework, which is now ready. Most of these UI changes are still experimental and are designed to be revertible without tremendous effort.

 

This is a 12.9" iPad Pro, trying to show how we can achieve big buttons (per design) without taking up the entire vertical space, which looked _really_ strange

https://user-images.githubusercontent.com/143916/111503641-4bc5fd00-871d-11eb-91fc-7b1e728a71f7.mov


Vocable/Common/PresetItemCollectionViewCell.swift
```diff
@@ -56,11 +56,14 @@ class PresetItemCollectionViewCell: VocableCollectionViewCell {
...
         rightConstraint.priority = .init(999)
+
+        let bottomConstraint = textLabel.bottomAnchor.constraint(equalTo: contentView.layoutMarginsGuide.bottomAnchor)
+        bottomConstraint.priority = .init(999)
         NSLayoutConstraint.activate([
             rightConstraint,
             textLabel.topAnchor.constraint(equalTo: contentView.layoutMarginsGuide.topAnchor),
             textLabel.leftAnchor.constraint(equalTo: contentView.layoutMarginsGuide.leftAnchor),
-            textLabel.bottomAnchor.constraint(equalTo: contentView.layoutMarginsGuide.bottomAnchor)
+            bottomConstraint
```
Clstroud, line 66: We previously only had a breakable constraint on the horizontal axis, which only covered _some_ of the potential constraint errors.

Vocable/Common/Views/BorderedView.swift
```diff
@@ -8,15 +8,31 @@
...
+                let newAction = basicAction.copy() as! CABasicAnimation
+                newAction.keyPath = event
+                newAction.fromValue = presentation()?.value(forKey: event)
+                action = newAction
+            }
+        }
+        return action
+    }
+}
+
```
Clstroud, line 26: This allows the bordered view to animate its path as part of a UIView animation. I don't believe this is being used in the current implementation, but it isn't bad to have this to make sure those views animate as one would expect.

Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -25,98 +25,58 @@ private class EmptyStateButton: GazeableButton {
...
-                #warning("Needs localization & Final copy")
-                title = "Speech services unavailable"
-            }
-            return NSAttributedString(string: title, attributes: [.font: UIFont.boldSystemFont(ofSize: 24), .foregroundColor: UIColor.defaultTextColor])
+protocol EmptyStateRepresentable {
+    var title: String { get }
+    var description: String? { get }
+    var buttonTitle: String? { get }
+    var image: UIImage? { get }
+}
```
Clstroud, line 33: "Listening Mode" as a concept was completely taking over the EmptyStateType, which didn't feel right. I'm not 100% confident about this switch, but it at least allows us to break the notion of emptiness into separate domains, which feels better than a single type littered with mega switch statements.

Vocable/Common/VocableViewController.swift
```diff
@@ -46,6 +46,7 @@ import UIKit
         super.viewWillAppear(animated)
         setViewMarginsForCurrentTraitCollection()
         installBackButtonIfNeeded()
+        view.layoutIfNeeded()
```
Clstroud, line 49: When installing the empty state in viewDidLoad, it would throw constraint errors because the readableContentGuide wasn't sized yet. No amount of breakable constraints were solving the problem, but forcing a layout pass after the margins had been updated for the current trait collection seems to do the right thing.

I'm doing it in viewWillAppear instead of _inside_ `setViewMarginsForCurrentTraitCollection` because I don't want a layout pass to necessarily be a side effect of calling `setViewMarginsForCurrentTraitCollection`.

Vocable/Features/Root/CategoryDetailViewController.swift
```diff
@@ -30,6 +30,7 @@ class CategoryDetailViewController: PagingCarouselViewController, NSFetchedResul
         let request: NSFetchRequest<Phrase> = Phrase.fetchRequest()
 
         if category.identifier == Category.Identifier.recents {
+            request.predicate = NSComparisonPredicate(\Phrase.lastSpokenDate, .notEqualTo, nil)
```
Clstroud, line 33: This was preventing recents from showing its empty state. A fresh installation had recent content, which was a huge red flag 😉

Vocable/Features/Voice/AudioPermissionPromptController.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  AudioPermissionPromptPresenter.swift
+//  AudioPermissionPromptController.swift
```
Clstroud, line 2: Rather than doing something protocol-driven that monkeyed with the states of view controllers (which felt wrong anyway), this encapsulates that logic as an object that those view controllers can instantiate and use how they see fit.

stevefosterwta (approved):
🐛 🗄️ 👍

PR #442, opened 2021-03-25
Debug interface + listen mode transition fixes

Clstroud:
* Ensures debug content is ordered by recency
* Fixes transitions for empty states so they remain in place while responses animate vertically


stevefosterwta (approved):
👍

PR #444, opened 2021-03-26
Hotword Enhancements

Clstroud:
Attempts to fix a few hotword-related issues:
* Invocation of the hotword would sometimes have a preceding syllable (usually something like a stray 'I'), which would be caused by any slight hesitation leading into the "hey." Rather than simply performing a find/replace on the hotword, this change will cause anything preceding the hotword to be removed from the transcription. This should be a safe change since the hotword is meant to be an antecedent to the actual phrase the speaker intends to be transcribed.
* ASR can be finicky, so this attempts to make hotword recognition more flexible by using a regex to encapsulate common ASR mistakes
* If the listen category was already selected, sometimes ASR would successfully hear "hey vocable" and strip it out, but it would leave "hey" as the transcription. This was because we didn't recognize partial transcriptions as being potential hotword candidates in intermediate transcriptions ("hey" shows up before "hey vocable"). This change now ensures that the regex for the first word that ASR _might_ hear is excluded from transcriptions until a subsequent utterance follows up with another word to ensure the transcription is _not_ the hotword.

### DIFF
--- a/Vocable.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Vocable.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/willowtreeapps/vocable-ios-spm.git",
         "state": {
           "branch": null,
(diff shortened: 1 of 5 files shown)


stevefosterwta (approved):
👍

PR #445, opened 2021-04-02
Listen mode numerical responses

Clstroud:
* Integrates a new model version that supports numerical questions (e.g. “How many …”)
* Resets the transcription when navigating away from the listening category to increase clarity


stevefosterwta (approved):
👍

Issue #405, opened 2020-09-17
Unlink the Chained UI Tests!


rc-young:
Currently, UI tests for iOS are dependent on each other, which causes issues if one were to stop working. We need to unlink and make each test independent of each other.

Creating a [Task List](https://docs.github.com/en/issues/tracking-your-work-with-issues/quickstart#adding-a-task-list) for sub-tasks of this parent ticket. We will update this list as it grows.

- [x] Unlink CustomCategoriesTests Class' tests #464 
    - [x] Re-enable tests; refactor to leverage `resetAtLaunch` build launch argument #511 
- [x] Handle the CustomCategoriesTests::testPagination test #470 
- [x] Fix the SettingScreenTests::testAddCustomCategory test #514  
- [x] Fix the KeyboardScreenTests #466 
- [x] Fix the SettingScreenTests #480
- [x] Re-enable and Fix the MainScreenTests #526  
- [x] Fix other (if any) failing tests
    - [x] SettingsScreenTests we disabled for the design changes on Settings > Categories screen

Issue #114, opened 2020-03-09
Reset App Settings


kyliefk:
**As a vocable app (iOS) user, I would like to be able to reset the app to the default settings**

**As a tester, when automating I would like to be able to trigger the reset functionality in my test code so that I do not need to randomize category names and phrases.**

**Designs**
NEEDED

For Automated Testing:
We would like to be able to trigger this reset at the launch of the app so we don't need to randomize our test data; e.g. category names. This should be an environment variable flag.

**Acceptance Criteria**
- Users can tap on the Settings > 'Reset App Settings' button to reset the app to the original default settings and content.
- An environment variable flag exists that can be set from within UI test code so we can reset the app data during testing.

rhsalas3:
@Clstroud 

Per our conversation about using random strings in test code. I have updated this ticket to include our solution; the environment variable flag.

PR #467, opened 2022-03-14
Unlink chained UI tests: CustomCategoriesTests

### LINKED ISSUE #464: Unlink CustomCategoriesTests Class' tests
Tests in the CustomCategoriesTests Class are all reliant on each other. These tests need to be able to run independently to reduce potential failure rate of the suite. 

This issue is a sub-task of #405

rhsalas3:
This PR closes #464 

The parent issue to #464 is #405

# Description of Work
Unlinked the tests defined in CustomCategoriesTests by creating a new base test Class for them; CustomCategoriesBaseTest. This new Class sets up a random custom category at the onset of each test.

Other work included refactors to the existing tests to account for new/changed flows and corrections to element queries.

------

## Notes to Test
All tests in CustomCategoriesTests should pass now. There is one test we left disabled. There is a sub-task in parent ticket #405 for addressing this test, and those like it.

### DIFF
--- a/Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
+++ b/Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
@@ -45,6 +45,7 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
             let button = GazeableButton()
             button.setImage(UIImage(systemName: "arrow.left"), for: .normal)
             button.addTarget(self, action: #selector(handleBackButton(_:)), for: .primaryActionTriggered)
+            button.accessibilityIdentifier = "navigationBar.backButton"
             return button
         }()
     }
@@ -167,6 +168,10 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
             if !category.isUserGenerated {
                 cell.editButton.isEnabled = false
             }
+            
+            // Assign identifiers for automation
+            cell.accessibilityIdentifier = "category_title"
+            cell.editButton.accessibilityIdentifier = "category_title_edit_button"
 
             return cell
 
@@ -179,6 +184,10 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
                 cell.showCategorySwitch.isOn = !category.isHidden
                 cell.showCategorySwitch.isEnabled = (category.identifier != .userFavorites)
             }
+            
+            // Assign an identifier for automation
+            cell.showCategorySwitch.accessibilityIdentifier = "show_category_toggle"
+            
             return cell
 
         case .addPhrase:
@@ -186,12 +195,20 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
             let title = NSLocalizedString("Edit Phrases", comment: "Edit Phrases")
             cell.setup(title: title, image: UIImage(systemName: "chevron.right"))
             cell.isEnabled = shouldEnableItem(at: indexPath)
+            
+            // Assign an identifier for automation
+            cell.accessibilityIdentifier = "edit_phrases_cell"
+            
             return cell
 
         case .removeCategory:
             let cell = collectionView.dequeueReusableCell(withReuseIdentifier: EditCategoryRemoveCollectionViewCell.reuseIdentifier, for: indexPath) as! EditCategoryRemoveCollectionViewCell
             cell.isEnabled = shouldEnableItem(at: indexPath)
             cell.textLabel.text = NSLocalizedString("category_editor.detail.button.remove_category.title", comment: "Remove category button label within the category detail screen.")
+            
+            // Assign an identifier for automation
+            cell.accessibilityIdentifier = "remove_category_cell"
+            
             return cell
         }
     }
--- a/VocableUITests/Screens/BaseScreen.swift
+++ b/VocableUITests/Screens/BaseScreen.swift
@@ -15,6 +15,26 @@ import XCTest
 class BaseScreen {
     let paginationLabel = XCUIApplication().staticTexts["bottomPagination.pageNumber"]
     
+    /// From Pagination: the current page (X) being viewed from the "Page X of Y" pagination label.
+    var currentPageNumber: Int {
+        // Define a regex pattern with named matching group, 'current', for reference
+        let pattern = #"(?<current>\d)+ .+ (\d)+"#
+        
+        let pageNumber = getNamedGroupInRegexPatternFromText(namedGroup: "current", regexPattern: pattern, text: paginationLabel.label)
+        
+        return Int(pageNumber)!
+    }
+    
+    /// From Pagination: the total number of pages (Y) from the "Page X of Y" pagination label.
+    var totalPageCount: Int {
+        // Define a regex pattern with named matching group, 'total', for reference
+        let pattern = #"(\d)+ .+ (?<total>\d)+"#
+        
+        let pageCount = getNamedGroupInRegexPatternFromText(namedGroup: "total", regexPattern: pattern, text: paginationLabel.label)
+        
+        return Int(pageCount)!
+    }
+    
     /**
      This method extracts a substring from text, that matches the named group within a defined regex pattern.
      
@@ -26,7 +46,7 @@ class BaseScreen {
     private func getNamedGroupInRegexPatternFromText(namedGroup: String, regexPattern: String, text: String) -> String {
         var matchingText = ""
         let regex = try! NSRegularExpression(pattern: regexPattern)
-        let range = NSRange(location: 0, length: text.count) // Used for the NSRegularExpression
+        let range = NSRange(text.startIndex..<text.endIndex, in: text) // Used for the NSRegularExpression
         
         // Find a match to the regex pattern...
         if let match = regex.firstMatch(in: text, options: [], range: range) {
@@ -38,24 +58,6 @@ class BaseScreen {
         }
         
         return matchingText
-    }
-    
-    /// For Pagination: retrieve the current page (X) being viewed
-    /// from the "Page X of Y" pagination label.
-    func getCurrentPage() -> String {
-        // Define a regex pattern with named matching group, 'current', for reference
-        let pattern = #"(?<current>\d)+ of (\d)+"#
-        
-        return getNamedGroupInRegexPatternFromText(namedGroup: "current", regexPattern: pattern, text: paginationLabel.label)
-    }
-    
-    /// For Pagination: retrieve the total number of pages (Y) from the
-    ///  "Page X of Y" pagination label.
-    func getTotalNumOfPages() -> String {
-        // Define a regex pattern with named matching group, 'total', for reference
-        let pattern = #"(\d)+ of (?<total>\d)+"#
-        
-        return getNamedGroupInRegexPatternFromText(namedGroup: "total", regexPattern: pattern, text: paginationLabel.label)
     }
     
 }
--- a/VocableUITests/Screens/BaseScreen.swift
+++ b/VocableUITests/Screens/BaseScreen.swift
@@ -13,51 +13,49 @@ import Foundation
 import XCTest
 
 class BaseScreen {
-    // Define a regex pattern with named matching groups, 'current' and 'total', for reference
-    let paginationRegexPattern = #"(?<current>\d)+ of (?<total>\d)+"#
-    
     let paginationLabel = XCUIApplication().staticTexts["bottomPagination.pageNumber"]
+    
+    /**
+     This method extracts a substring from  text, that matches the named group within a defined regex pattern.
+     
+     For regex pattern #"Find (?\<groupdName\>\w)+ please"#, the \<groupName\> is matched from 'Find help please'
+     to return 'help'
+     
+     Documentaion: https://nshipster.com/swift-regular-expressions/
+    */
+    private func getNamedGroupInRegexPatternFromText(namedGroup: String, regexPattern: String, fromText: String) -> String {
+        var matchingText = ""
+        let regex = try! NSRegularExpression(pattern: regexPattern)
+        let range = NSRange(location: 0, length: fromText.count) // Used for the NSRegularExpression
+        
+        // Find a match to the regex pattern...
+        if let match = regex.firstMatch(in: fromText, options: [], range: range) {
+            // Within the match to the regex pattern, extract the matching group by its name
+            let resultingMatch = match.range(withName: namedGroup)
+
+            // Extract the Substring (as a String) of the found match from the range returned by regex.firstMatch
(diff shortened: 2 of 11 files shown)


VocableUITests/Tests/CustomCategoriesTests.swift
```diff
@@ -63,63 +59,48 @@ class CustomCategoriesTest: BaseTest {
...
-        
+        // Assert that now we have two cells containing the same phrase
+        let phrasePredicate = NSPredicate(format: "label MATCHES %@", testPhrase)
+        let phraseQuery = XCUIApplication().staticTexts.containing(phrasePredicate)
+        XCTAssertEqual(phraseQuery.count, 2, "Expected both phrases to be present")
     }
     
+    // TODO: Disabled for now. Moving it to a different test class as part of issue #405
+    // https://github.com/willowtreeapps/vocable-ios/issues/405
     func testPagination() {
```
SashankPatel, line 104: Did you mean to disable/comment out this test?

rhsalas3: Yes. I'm working on that test and the other pagination tests in a different issue (which I haven't written up yet). I have it as a task in issue #405 . 

Sorry, I should have made that more clear in my comment.

Update: I just created an issue/ticket for addressing this test, #470

SashankPatel: Spoke with Rudy, the one test that is failing is being worked in a separate task.  Approving this PR with that caveat.

VocableUITests/Screens/BaseScreen.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+    
+    /// For Pagination: retrieve the total number of pages (Y) from the
+    ///  "Page X of Y" pagination label.
+    func getTotalNumOfPages() -> String {
+        // Define a regex pattern with named matching group, 'total', for reference
+        let pattern = #"(\d)+ of (?<total>\d)+"#
+        
+        return getNamedGroupInRegexPatternFromText(namedGroup: "total", regexPattern: pattern, text: paginationLabel.label)
+    }
```
Clstroud: Minor Swift-isms: 
1. I wouldn't usually expect to see a "getter" function such as these in Swift. They may read better as computed properties? e.g. `var currentPageNumber: String { ... }`
2. Abbreviated names can often be switched up for clarity e.g. `getTotalNumOfPages` -> `currentTotalPageCount`
3. If I'm reading ahead correctly, these are being converted to `Int`s later, so perhaps it makes sense to have them return `Int`s here instead of `String`s?

VocableUITests/Screens/CustomCategoriesScreen.swift
```diff
@@ -8,13 +8,14 @@
...
 
-class CustomCategoriesScreen {
+class CustomCategoriesScreen: BaseScreen {
 
     let settingsScreen = SettingsScreen()
     let keyboardScreen = KeyboardScreen()
     let mainScreen = MainScreen()
     
     let categoriesPageAddPhraseButton = XCUIApplication().buttons["settingsCategory.addPhraseButton"]
+    let editCategoryPhrasesCell = XCUIApplication().cells.containing(.staticText, identifier: "Edit Phrases").element
```
Clstroud: Should we make a dedicated AX identifier for this cell in case the localization changes?

VocableUITests/Tests/CustomCategoriesBaseTest.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+//  Created by Rudy Salas on 3/10/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import XCTest
+
+class CustomCategoriesBaseTest: BaseTest {
+    
+    var customCategory: String = ""
```
Clstroud: As a third party I'm not immediately sure what about the category this is referencing. Would a rename to `customCategoryName` make sense?

Clstroud: Is this property being mutated by subclasses? Looks like maybe it could be declared as `private(set)` to make it clear that you're supposed to use `setCustomCategory(...` to change it?

VocableUITests/Tests/CustomCategoriesBaseTest.swift
```diff
@@ -0,0 +1,33 @@
...
+        
+        setCustomCategory(name: "Hi", numOfRandomLetters: 3)
+        
+        // Create a custom category and open it
+        settingsScreen.navigateToSettingsCategoryScreen()
+        customCategoriesScreen.createCustomCategory(categoryName: customCategory)
+        settingsScreen.openCategorySettings(category: customCategory)
+    }
+    
+    func setCustomCategory(name: String, numOfRandomLetters: Int) {
```
Clstroud: I'm curious about the random letters thing: is the goal here to avoid collisions with other categories? Is that better than having hardcoded suffixes for each test to guarantee there can't be collisions?

Not saying this is wrong, just thinking about the determinism angle 🤔

VocableUITests/Tests/CustomCategoriesBaseTest.swift
```diff
@@ -0,0 +1,33 @@
...
+        customCategory = name + randomString(length: numOfRandomLetters).lowercased()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+    }
```
Clstroud, line 32:
```suggestion
```
Doesn't look like this is doing anything

Clstroud (commented):
Solid improvements 💯 I left a few notes that we can discuss if you want. Most are minor and some may not be worth the effort.

VocableUITests/Tests/CustomCategoriesBaseTest.swift
rhsalas3: Gotcha. I wasn't sure if the super class' tearDown() had been called during some of the test runs so I added it to the new base class to be sure. 

I will remove it.

VocableUITests/Tests/CustomCategoriesBaseTest.swift
rhsalas3: I've never used suffixes in test names to guarantee test order. Is that what you're suggesting? 

Your assumption is correct. The purpose is to randomize the name so that subsequent runs of the same test or other tests at the same time do not conflict. For context: each test was originally written to piggy back off of one another but that is problematic if one of the tests fails ahead of others downstream.

VocableUITests/Tests/CustomCategoriesBaseTest.swift
rhsalas3: This property was intended to be shared test data for the subclasses, but not mutable. I wanted a base class for the tests in CustomCategoriesTests that would create a category as the setup for each test.  So I wanted to be able to just reference the category name from the super class.

The `setCustomCategory()` function is, currently, intended to be used only w/in to the CustomCategoriesBaseTest. So maybe I should make the function private. Thoughts?

VocableUITests/Screens/CustomCategoriesScreen.swift
rhsalas3: At the time of writing this, I didn't know enough about the code base to assign a dedicated a11y identifier lol Now that we have had the conversation about the category title cell, I think I can handle adding one for this element.

I will make those changes.

VocableUITests/Screens/BaseScreen.swift
```diff
@@ -0,0 +1,61 @@
...
+     
+     For regex pattern #"Find (?\<groupdName\>\w)+ please"#, the \<groupName\> is matched from 'Find help please'
+     to return 'help'
+     
+     Documentaion: https://nshipster.com/swift-regular-expressions/
+    */
+    private func getNamedGroupInRegexPatternFromText(namedGroup: String, regexPattern: String, text: String) -> String {
+        var matchingText = ""
+        let regex = try! NSRegularExpression(pattern: regexPattern)
+        let range = NSRange(location: 0, length: text.count) // Used for the NSRegularExpression
```
moyerr: When dealing with Swift Strings, it is recommended to avoid this initializer and instead use the `NSRange(_:in:)` initializer, which accepts a Swift `Range` and the source string. The reason is that in Swift, String indices are not simply integer offsets (because they have to account for characters that comprise multiple unicode scalars), so a range spanning `0` to `text.count` may not necessarily be equal to a range spanning `text.startIndex` to `text.endIndex` and therefore wouldn't necessarily capture the entire string.

The issue hasn't presented itself in the use cases so far, but this is a useful piece of code and the range discrepancy could be a gnarly bug to track down if this gets copied or reused in more places down the road 🙂

Feel free to reach out if you have any questions or need clarification!

VocableUITests/Screens/BaseScreen.swift
```diff
@@ -0,0 +1,61 @@
...
+        }
+        
+        return matchingText
+    }
+    
+    /// For Pagination: retrieve the current page (X) being viewed
+    /// from the "Page X of Y" pagination label.
+    func getCurrentPage() -> String {
+        // Define a regex pattern with named matching group, 'current', for reference
+        let pattern = #"(?<current>\d)+ of (\d)+"#
```
moyerr: Love the use of the extended string delimiters for the regex pattern 👏

VocableUITests/Screens/BaseScreen.swift
rhsalas3: RE:

1. Sure! That might be a cool "swift-ism" to implement. I'll give making them computed properties a try.
2. Gotcha. I will implement better names; something like `currentPageNumber` and `totalPageCount`
3. I thought about that and decided that since I was extracting the value out of a String then I would return it as the same type. Usually with test code, I like to manipulate my data w/in the test instead of within the helper functions for my final purpose. The future tests I intend these functions for would be referencing the labels, so I only needed to convert the value for the loop. But I might need them as `Int`s again later lol 🤔 ...I think returning as `Int`s now makes sense after thinking out loud. If anything because I've only had the one use for the value at the moment, as an `Int`.

I'll make these changes.

Clstroud: As for #3 I think I get what you mean, and that's a totally valid perspective if that's how you like to approach things! My only pushback would probably be naming. In the context of your initial implementation I would be surprised if a function I tried to call named `getTotalNumOfPages()` gave me back a `String` instead of a "Num" 😉

VocableUITests/Screens/CustomCategoriesScreen.swift
Clstroud: If you ever get a tingly feeling that you'd like an AX identifier, absolutely reach out for guidance on how to achieve it. We can make it happen 😎

VocableUITests/Tests/CustomCategoriesBaseTest.swift
Clstroud: 1. Sounds like declaring it as e.g. `private(set) var customCategoryName: String = ""` would achieve the "subclasses can see it but not mutate it" goal
2. If `CustomCategoriesBaseTest` is the only caller for `setCustomCategory()` then I agree -- make that sucker private

VocableUITests/Tests/CustomCategoriesBaseTest.swift
Clstroud: I'm not suggesting it for test order so much as I'm (gently) questioning the practice of relying on random strings to create unique category names. Since it's not something you explicitly control, you could still create a flakier test vs something that is fully deterministic. 

If you wanted a unique category name for each test for example, you could do something like 
```
    private func test_createCustomCategoryName() {
        let categoryName = "TestCategory" + #function
        print(categoryName) // TestCategorytest_createCustomCategoryName()
    }
```

VocableUITests/Screens/BaseScreen.swift
Clstroud: Timely and somewhat relevant discussion that I meant to post in the WT Swift channel: https://forums.swift.org/t/string-index-unification-vs-bidirectionalcollection-requirements/55946

VocableUITests/Screens/BaseScreen.swift
rhsalas3: That makes complete sense. Naming is very important.

VocableUITests/Screens/CustomCategoriesScreen.swift
rhsalas3: 🙏  Thank you! 🙏

VocableUITests/Screens/BaseScreen.swift
rhsalas3: Good to know @moyerr!

I will implement this change.

This also made me realize that my expression is only good for English because of the 'of' literal. I am going to update that to work with other languages.

Maybe further down the road the team and I can review these UI tests as a whole with regards to localization.

VocableUITests/Screens/BaseScreen.swift
rhsalas3: Changes implemented and pushed up!

VocableUITests/Tests/CustomCategoriesBaseTest.swift
rhsalas3: Changes implemented and pushed up!

VocableUITests/Screens/BaseScreen.swift
rhsalas3: Changes implemented and pushed up!

VocableUITests/Tests/CustomCategoriesBaseTest.swift
rhsalas3: Per a conversation with @Clstroud  on Slack:

For the issue of needing a unique category name each time we run a test, we can actually trigger the 'Reset App Settings' functionality at app launch. This will ensure that any of our previous custom categories and phrases are removed. Once implemented, we can create an environment variable flag that we can use w/in our tests to trigger the reset when necessary. Issue #114 will cover the dev work to implement the reset feature and env flag.

For now, we won't need to make any changes to this code around the use of the random strings.

Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
```diff
@@ -45,6 +45,7 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
             let button = GazeableButton()
             button.setImage(UIImage(systemName: "arrow.left"), for: .normal)
             button.addTarget(self, action: #selector(handleBackButton(_:)), for: .primaryActionTriggered)
+            button.accessibilityIdentifier = "navigationBar.backButton"
```
Clstroud, line 48: Loving all these IDs

Clstroud (approved):
💯 💰 Awesome job @rhsalas3

PR #471, opened 2022-03-15
#465 Allow categories and phrases to be universally editable

### LINKED ISSUE #465: Make categories and phrases universally editable
Using the existing UI implementation (for now), a user should be able to *rename* and *remove* all categories and phrases. The current implementation restricts these actions to custom, user-defined categories and phrases.

Clstroud:
Closes #465

# Description of Work

### Functionally
If the category or phrase is a preset, it can be renamed and deleted just like their user-defined counterparts
- Renaming means marking them as "user-renamed" so that subsequent updates of the presets don't overwrite the user's entry
- During "deletion" the category or phrase is marked as "user-removed" so that subsequent updates to the presets don't recreate the removed entity. We simply suppress it from showing in the UI at all.
- User-generated categories and phrases are still permanently deleted because there is no consequence for that

### Code changes
- I felt that the way Core Data predicates were being handled was getting a bit terse, so I made some lightweight global functions and operators to make dealing with them more readable and concise. I steered away from making anything too fancy (I hope), so hopefully these strike a nice balance.

## Notes to Test 
- This does not change the UI from previous builds beyond enabling/disabling buttons
- The "My Sayings" category is still a special case that is not deletable 
- We should do some testing to ensure we upgrade gracefully from the App Store version of the app
   - Specifically the "My Sayings" category migration

### DIFF
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -159,11 +159,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         let request: NSFetchRequest<Phrase> = Phrase.fetchRequest()
         request.predicate = {
-            let identifiers = Set(presets.phrases.map { $0.id })
-            let isNotUserGenerated = NSComparisonPredicate(\Phrase.isUserGenerated, .equalTo, false)
-            let identifierInSet = NSComparisonPredicate(\Phrase.identifier, .in, identifiers)
-            let identifierNotInSet = NSCompoundPredicate(notPredicateWithSubpredicate: identifierInSet)
-            return NSCompoundPredicate(andPredicateWithSubpredicates: [isNotUserGenerated, identifierNotInSet])
+            let identifiers = Set(presets.phrases.map(\.id))
+            let isNotUserGenerated = !Predicate(\Phrase.isUserGenerated)
+            let identifierNotInSet = !Predicate(\Phrase.identifier, .in, identifiers)
+            return isNotUserGenerated && identifierNotInSet
         }()
 
         let results = try context.fetch(request)
@@ -177,10 +176,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let request: NSFetchRequest<Category> = Category.fetchRequest()
         request.predicate = {
             let identifiers = Set(presets.categories.map { $0.id })
-            let isNotUserGenerated = NSComparisonPredicate(\Category.isUserGenerated, .equalTo, false)
-            let identifierInSet = NSComparisonPredicate(\Category.identifier, .in, identifiers)
-            let identifierNotInSet = NSCompoundPredicate(notPredicateWithSubpredicate: identifierInSet)
-            return NSCompoundPredicate(andPredicateWithSubpredicates: [isNotUserGenerated, identifierNotInSet])
+            let isNotUserGenerated = !Predicate(\Category.isUserGenerated)
+            let identifierNotInSet = !Predicate(\Category.identifier, .in, identifiers)
+            return isNotUserGenerated && identifierNotInSet
         }()
 
         let results = try context.fetch(request)
@@ -195,10 +193,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             // Legacy favorites category was .isUserGenerated = true with identifier = localized version
             // of "My Sayings." Going forward, all identifiers for Phrases/Categories are prefixed, so
             // we can use that to isolate the legacy category entry
-            let isUserGenerated = NSComparisonPredicate(\Category.isUserGenerated, .equalTo, true)
-            let identifierPrefixed = NSComparisonPredicate(\Category.identifier, .beginsWith, "user_")
-            let identifierNotPrefixed = NSCompoundPredicate(notPredicateWithSubpredicate: identifierPrefixed)
-            return NSCompoundPredicate(andPredicateWithSubpredicates: [isUserGenerated, identifierNotPrefixed])
+            let isUserGenerated = Predicate(\Category.isUserGenerated)
+            let identifierNotPrefixed = !Predicate(\Category.identifier, .beginsWith, "user_")
+            return isUserGenerated && identifierNotPrefixed
         }()
 
         let results = try context.fetch(request)
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -159,10 +159,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         let request: NSFetchRequest<Phrase> = Phrase.fetchRequest()
         request.predicate = {
-            let identifiers = Set(presets.phrases.map(\.id))
+            let presetPhraseIdentifiers = Set(presets.phrases.map(\.id))
             let isNotUserGenerated = !Predicate(\Phrase.isUserGenerated)
-            let identifierNotInSet = !Predicate(\Phrase.identifier, .in, identifiers)
-            return isNotUserGenerated && identifierNotInSet
+            let isNotPresetPhrase = !Predicate(\Phrase.identifier, isContainedIn: presetPhraseIdentifiers)
+            return isNotUserGenerated && isNotPresetPhrase
         }()
 
         let results = try context.fetch(request)
@@ -175,10 +175,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         let request: NSFetchRequest<Category> = Category.fetchRequest()
         request.predicate = {
-            let identifiers = Set(presets.categories.map { $0.id })
+            let presetCategoryIdentifiers = Set(presets.categories.map(\.id))
             let isNotUserGenerated = !Predicate(\Category.isUserGenerated)
-            let identifierNotInSet = !Predicate(\Category.identifier, .in, identifiers)
-            return isNotUserGenerated && identifierNotInSet
+            let isNotPresetCategory = !Predicate(\Category.identifier, isContainedIn: presetCategoryIdentifiers)
+            return isNotUserGenerated && isNotPresetCategory
         }()
 
         let results = try context.fetch(request)
@@ -193,9 +193,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             // Legacy favorites category was .isUserGenerated = true with identifier = localized version
             // of "My Sayings." Going forward, all identifiers for Phrases/Categories are prefixed, so
             // we can use that to isolate the legacy category entry
-            let isUserGenerated = Predicate(\Category.isUserGenerated)
-            let identifierNotPrefixed = !Predicate(\Category.identifier, .beginsWith, "user_")
-            return isUserGenerated && identifierNotPrefixed
+            let isUserGeneratedCategory = Predicate(\Category.isUserGenerated)
+            let isNotUserPrefixed = !Predicate(\Category.identifier, beginsWith: "user_")
+            return isUserGeneratedCategory && isNotUserPrefixed
         }()
 
         let results = try context.fetch(request)
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -218,8 +218,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private func updateDefaultCategories(in context: NSManagedObjectContext, withPresets presets: PresetData) throws {
         for presetCategory in presets.categories {
             let category = Category.fetchOrCreate(in: context, matching: presetCategory.id)
-            category.name = presetCategory.utterance
-            category.languageCode = presetCategory.languageCode
+            if !category.isUserRenamed {
+                category.name = presetCategory.utterance
+                category.languageCode = presetCategory.languageCode
+            }
             if category.isInserted {
                 category.isHidden = presetCategory.hidden
             }
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -227,9 +227,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         for presetPhrase in presets.phrases {
 
             let phrase = Phrase.fetchOrCreate(in: context, matching: presetPhrase.id)
-            phrase.utterance = presetPhrase.utterance
-            phrase.languageCode = presetPhrase.languageCode
-            
+            if !phrase.isUserRenamed {
+                phrase.utterance = presetPhrase.utterance
+                phrase.languageCode = presetPhrase.languageCode
+            }
             for identifier in presetPhrase.categoryIds {
                 if let category = Category.fetchObject(in: context, matching: identifier) {
                     phrase.category = category
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -208,8 +208,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         try updateDefaultCategories(in: context, withPresets: presets)
         try updateDefaultPhrases(in: context, withPresets: presets)
-        try updateCategoryForUserGeneratedPhrases(in: context)
-
     }
 
     private func updateDefaultCategories(in context: NSManagedObjectContext, withPresets presets: PresetData) throws {
@@ -238,22 +236,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                     category.addToPhrases(phrase)
                 }
             }
-        }
-    }
-
-    private func updateCategoryForUserGeneratedPhrases(in context: NSManagedObjectContext) throws {
-        let mySayingsCategory = Category.fetch(.userFavorites, in: context)
-        let request: NSFetchRequest<Phrase> = Phrase.fetchRequest()
-
-        let firstPredicate = NSComparisonPredicate(\Phrase.isUserGenerated, .equalTo, true)
-        let secondPredicate = NSComparisonPredicate(\Phrase.category?.isUserGenerated, .equalTo, false)
-        request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [firstPredicate, secondPredicate])
-
-        let phraseResults = try context.fetch(request)
-
-        for phrase in phraseResults {
-            phrase.category = mySayingsCategory
-            mySayingsCategory.addToPhrases(phrase)
         }
     }
 }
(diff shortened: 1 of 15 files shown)


Vocable/Extensions/PredicateBuilders.swift
```diff
@@ -0,0 +1,177 @@
...
+
+typealias PredicateOperator = NSComparisonPredicate.Operator
+
+/// Constructs an `NSPredicate` where the given `KeyPath` must equal `true`
+///
+/// - Parameter lhs: `KeyPath` with a `Bool`value that is used for comparison
+/// - Returns: `NSPredicate`
+func Predicate<A>(_ lhs: KeyPath<A, Bool>) -> NSPredicate {
+    Predicate(lhs, equalTo: true)
+}
```
moyerr, line 19: Using global functions with capitalized names really provides the ergonomics of dealing with a more concise and type-safe wrapper type, but without the burden of bridging with all the CoreData APIs. I like it! 

I don't think I've seen this pattern before; was there a particular inspiration to go this direction here? If I was doing the same thing, I probably would've reached for a `typealias` with a series of convenience initializers. I don't think one is necessarily better than the other, I'm just curious what led to this particular design choice 🙂

Clstroud: Thanks! Glad it's as intuitive as I'd hoped it might be 🙂 You're spot-on about the goal of the ergonomics.

As for inspiration: I wasn't modeling it off of anything in particular. I kicked around a couple ideas that didn't quite feel right and eventually landed here. Part of the influence was the [existing code](https://github.com/willowtreeapps/vocable-ios/blob/74fe115b62e81dadbaf96e92c88654846ec87881/Vocable/CoreData/NSManagedObject%2BHelpers.swift#L52) that wasn't really cutting it.

I think having a `typealias Predicate` would be a sensible starting point for sure, but you'd still be confined to a particular class of `NSPredicate` and unable to choose which subclass you want to vend. I believe all of these are vending an `NSComparisonPredicate`, but I can imagine a few additional functions we could add that might want to switch it up and vend an `NSCompoundPredicate` instead. Can't really do that with a custom initializer.

Moreover, the convenience logical operators I added inherently work more or less the same way, so there's something to be said about the consistency of `-> NSPredicate`.

Almost like everything is returning an `any Predicate` without the existential dread 🥁

Clstroud: Other thoughts for future tinkering:
1. If we really wanted to get crazy, we could introduce operators that could make e.g. `Predicate(\Category.isUserGenerated == false)` valid
2. Potential ergonomic improvements by making these an extension on `NSManagedObject` e.g. `Category.Predicate(\.isUserGenerated == false)` but that feels a little more subjective ¯\_(ツ)_/¯

Vocable/AppDelegate.swift
```diff
@@ -195,10 +193,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             // Legacy favorites category was .isUserGenerated = true with identifier = localized version
             // of "My Sayings." Going forward, all identifiers for Phrases/Categories are prefixed, so
             // we can use that to isolate the legacy category entry
-            let isUserGenerated = NSComparisonPredicate(\Category.isUserGenerated, .equalTo, true)
-            let identifierPrefixed = NSComparisonPredicate(\Category.identifier, .beginsWith, "user_")
-            let identifierNotPrefixed = NSCompoundPredicate(notPredicateWithSubpredicate: identifierPrefixed)
-            return NSCompoundPredicate(andPredicateWithSubpredicates: [isUserGenerated, identifierNotPrefixed])
+            let isUserGenerated = Predicate(\Category.isUserGenerated)
+            let identifierNotPrefixed = !Predicate(\Category.identifier, .beginsWith, "user_")
+            return isUserGenerated && identifierNotPrefixed
```
jsmorgan42: Definitely a fan of these changes, makes it much easier to reason what's happening here 👏

Vocable/AppDelegate.swift
```diff
@@ -241,20 +239,4 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-        let secondPredicate = NSComparisonPredicate(\Phrase.category?.isUserGenerated, .equalTo, false)
-        request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [firstPredicate, secondPredicate])
-
-        let phraseResults = try context.fetch(request)
-
-        for phrase in phraseResults {
-            phrase.category = mySayingsCategory
-            mySayingsCategory.addToPhrases(phrase)
-        }
-    }
```
jsmorgan42, line 259: Is this basically adding all user generated phrases to the my sayings category? Is that why this was removed?

Vocable/CoreData/Phrases.xcdatamodeld/Phrases v4.xcdatamodel/contents
```diff
@@ -0,0 +1,30 @@
...
+        <attribute name="isUserGenerated" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="isUserRemoved" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="isUserRenamed" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="languageCode" optional="YES" attributeType="String"/>
+        <attribute name="name" attributeType="String"/>
+        <attribute name="ordinal" attributeType="Integer 32" defaultValueString="0" usesScalarValueType="YES"/>
+        <relationship name="phrases" toMany="YES" deletionRule="Cascade" destinationEntity="Phrase" inverseName="category" inverseEntity="Phrase"/>
+    </entity>
+    <entity name="Phrase" representedClassName="Phrase" syncable="YES" codeGenerationType="class">
+        <attribute name="creationDate" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
```
jsmorgan42, line 16: Was curious why the creation date for a phrase is optional?

SashankPatel:
Verified phrases in preset categories can now be removed from preset categories and empty state appears.
Verified Selected category on the main screen can be deleted.

ISSUE(RESOLVED): Preset Categories can not be removed. Custom Categories can be removed.
- Verified preset categories are removed.

Vocable/Extensions/PredicateBuilders.swift
jsmorgan42: @Clstroud I like the first potential improvement a lot even though it might be slightly overkill. It feels like that would make things even more readable.

Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
```diff
@@ -234,9 +227,9 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
         case .showCategoryToggle:
             return (category.identifier != .userFavorites)
         case .addPhrase:
-            return (category.identifier == Category.Identifier.userFavorites) || category.isUserGenerated
+            return true
```
jsmorgan42: Can we just combine this into one case with `.titleEditView` on line 225 since they both return true?

Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
```diff
@@ -294,38 +287,55 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
...
-        self.navigationController?.popViewController(animated: true)
+        guard let category = category else { return }
+        if category.isUserGenerated {
+            context.delete(category)
+        } else {
+            category.isUserRemoved = true
+        }
+        if saveContext() {
+            self.navigationController?.popViewController(animated: true)
+        }
```
jsmorgan42, line 316: This can definitely be cleared up in another piece of work, but I wonder if we should show some kind of error alert if the category fails to get removed? Or are we just pretty confident it would only ever fail if something is real messed up?

Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
```diff
@@ -294,38 +287,55 @@ final class EditCategoryDetailViewController: VocableCollectionViewController, E
...
         } catch {
             assertionFailure("Failed to unsave user generated phrase: \(error)")
         }
+        return false
     }
 
     // MARK: EditCategoryDetailTitleCollectionViewCellDelegate
 
     func didTapEdit() {
+
```
jsmorgan42: Extra space

Vocable/AppDelegate.swift
Clstroud: Yep! We may need to bring this back in _some_ form once we test the App Store upgrade path, but to the very best of my knowledge this was just to handle upgrades from earlier iterations of the feature that were never released.

Vocable/CoreData/Phrases.xcdatamodeld/Phrases v4.xcdatamodel/contents
Clstroud: "Optional" in the context of Core Data is more about validation requirements when saving objects than what I assume you're probably think of (Swift.Optional). It's just marked optional here because that's the default in the editor, but it doesn't have much bearing on how we're using it

Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
Clstroud: If you wouldn't mind using the PR suggestions feature for this in the future, that'd be great 💯 

But yes, I'll take care of it.

Vocable/Features/Settings/EditCategories/EditCategoryDetailViewController.swift
Clstroud: Yeah, I mostly shifted it this way to ensure we couldn't just lose the user's entry when there's a failure, but for the moment I'm reasonably confident this will be fine. Having more robust error handling for these edits could be its own ticket honestly.

Clstroud:
Thanks to @rowildcard for finding the issue where preset category deletions were not being reflected by the UI! 

It led to a handful of other observations which I just pushed fixes for:

* Ensure synthesized categories can't have phrases edited (Numpad, Recents, Listening Mode)
* Ensure the initially-selected category on the main screen is not one that has been removed by the user
* Ensure that the reordering controls on the Settings -> Edit Categories screen were updated properly after a category is deleted (some were being disabled erroneously)

SashankPatel (changes requested):
Issue 1(RESOLVED): App crashes when "Star'ing" a phrase from the keyboard i.e. tapping the favorite button
Screen Shot 2022-03-17 at 8 05 08 AM
.


Issue 2: this is pretty minor: 
1.) Launch the app fresh so you have all the categories and the General Category should be visible.
2.) Navigate to Setting and delete the General Category. -> Basic Needs is now 1st in the list.
3.) Navigate back to main screen 
Expected: Basic needs to be visible.
Actual: Personal Care was visible.


Issue 1: has been resolved with latest commit.  Favorites can now be added to My Sayings
Issue 2: We decided to break this out into a separate bug ticket.

Vocable/AppDelegate.swift
```diff
@@ -159,11 +159,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         request.predicate = {
-            let identifiers = Set(presets.phrases.map { $0.id })
-            let isNotUserGenerated = NSComparisonPredicate(\Phrase.isUserGenerated, .equalTo, false)
-            let identifierInSet = NSComparisonPredicate(\Phrase.identifier, .in, identifiers)
-            let identifierNotInSet = NSCompoundPredicate(notPredicateWithSubpredicate: identifierInSet)
-            return NSCompoundPredicate(andPredicateWithSubpredicates: [isNotUserGenerated, identifierNotInSet])
+            let identifiers = Set(presets.phrases.map(\.id))
+            let isNotUserGenerated = !Predicate(\Phrase.isUserGenerated)
+            let identifierNotInSet = !Predicate(\Phrase.identifier, .in, identifiers)
+            return isNotUserGenerated && identifierNotInSet
```
moyerr: I wonder if we could give this Predicate a little more semantic meaning as it relates to the fetchRequest? If I'm understanding the set of identifiers correctly, maybe something like `identifierDoesNotMatchPreset`? Or if that's too verbose, maybe do the negation on the following line: `return isNotUserGenerated && !identifierMatchesPreset`

Vocable/AppDelegate.swift
```diff
@@ -177,10 +176,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let request: NSFetchRequest<Category> = Category.fetchRequest()
         request.predicate = {
             let identifiers = Set(presets.categories.map { $0.id })
```
moyerr: You updated the closure to a keypath above. Could we do the same here?

```suggestion
            let identifiers = Set(presets.categories.map(\.id))
```

Vocable/Extensions/NSPredicate+Operators.swift
```diff
@@ -0,0 +1,32 @@
...
+        NSCompoundPredicate(notPredicateWithSubpredicate: lhs)
+    }
+
+    public static func &= (_ lhs: inout NSPredicate, _ rhs: NSPredicate) {
+        lhs = NSCompoundPredicate(andPredicateWithSubpredicates: [lhs, rhs])
+    }
+
+    public static func |= (_ lhs: inout NSPredicate, _ rhs: NSPredicate) {
+        lhs = NSCompoundPredicate(orPredicateWithSubpredicates: [lhs, rhs])
+    }
```
moyerr: Do these need to be `public`?

Vocable/Extensions/PredicateBuilders.swift
```diff
@@ -0,0 +1,177 @@
...
+
+/// Constructs an `NSPredicate` where the given `KeyPath` must equal the given value
+///
+/// - Parameters:
+///   - lhs: `KeyPath` whose value must equal `rhs`
+///   - rhs: optional value which `lhs` must equal
+/// - Returns: `NSPredicate`
+func Predicate<A, B>(_ lhs: KeyPath<A, B?>, equalTo rhs: B?) -> NSPredicate {
+    Predicate(lhs, .equalTo, rhs)
+}
```
moyerr, line 62: Is it necessary to have an overload specifically for Optionals? Would these types not be covered by the more general version?

Vocable/Extensions/PredicateBuilders.swift
```diff
@@ -0,0 +1,177 @@
...
+/// - Parameters:
+///   - lhs: `KeyPath` whose value is compared against `rhs`
+///   - rhs: `Collection` which `lhs` will be compared against
+///   - operation: the method of comparison
+/// - Returns: `NSPredicate`
+func Predicate<A, B, C: Collection>(_ lhs: KeyPath<A, B>, _ operation: PredicateOperator, _ rhs: C) -> NSPredicate where C.Element == B {
+    let lhs = NSExpression(forKeyPath: lhs)
+    let rhs = NSExpression(forConstantValue: rhs)
+    return NSComparisonPredicate(leftExpression: lhs, rightExpression: rhs, modifier: .direct, type: operation)
+}
```
moyerr: I love the way the `Predicate(_:equalTo:)` and `Predicate(_:notEqualTo:)` functions really provide that ✨ _clarity at the point of use_ ✨. Do you think it's worth it to provide something similar for collection-specific operators so we can keep the `PredicateOperator` as an implementation detail? Something like:
- `Predicate(_:beginsWith:)`
- `Predicate(_:endsWith:)`
- `Predicate(_:isIn:)`

Vocable/Features/Root/RootEditTextViewController.swift
```diff
@@ -18,7 +18,6 @@ class RootEditTextViewController: EditTextViewController {
         let button = GazeableButton()
         button.setImage(UIImage(systemName: "star"), for: .normal)
         button.accessibilityIdentifier = "keyboard.favoriteButton"
-        button.addTarget(self, action: #selector(favoriteButtonSelected), for: .primaryActionTriggered)
```
moyerr, line 21: I think adding the target in `viewDidLoad` is perfectly valid, but an alternate approach I've used in the past is to make the view a `lazy var`. It's actually the [SwiftLint suggestion for this common mistake](https://realm.github.io/SwiftLint/self_in_property_initialization.html).

rowildcard:
Test verified. See issue for test details.

rhsalas3:
> Test verified. See issue for test details.

Can confirm: I tested and verified the expected behavior as well, including adding phrases to 'My Sayings' from the keyboard screen. 

These changes look good to me.

Vocable/Extensions/PredicateBuilders.swift
Clstroud: Yeah, this is a weird thing that I'm _pretty_ sure is necessary (but I could be proven wrong). Reason is: some values we treat as non-optional but in Core Data land just about everything is technically nullable and the interfaces of the NSManagedObjects reflect that. Basically we need this optionality for pattern-matching to succeed in some cases but I don't have an example immediately available. I could be misremembering exactly why this was necessary, or we may be able to express it differently, but that's what I've got as an answer at the moment 😅 

Always happy to yeet some unnecessary code if we find it's extraneous... but for the moment I'd like to get this sucker merged so we can keep rolling.

Vocable/Features/Root/RootEditTextViewController.swift
Clstroud: I have done that in plenty of situations as well, but this time I talked myself out of it. There's just something about the way `lazy` changes the meaning of `self` that strikes me as unintuitive for newer developers that pushed me to move it to `viewDidLoad()`, you know? Give me 24 hours and I'm sure I can be persuaded back in the other direction 😛

Vocable/Extensions/PredicateBuilders.swift
Clstroud: I will add just as many as are needed to make `PredicateOperator` a private implementation detail. And then that's all! 😉

Vocable/Extensions/PredicateBuilders.swift
moyerr: Yeah, I haven't dealt with `NSExpressions` or `NSPredicates` too much so I don't have enough context to know for sure. I'm good with it as-is though!

Vocable/Features/Root/RootEditTextViewController.swift
moyerr: No change needed! I'm good either way 💯

moyerr (approved):
Huge fan of the documentation comments and the descriptive commits that reference the issue! 🏆

Issue #466, opened 2022-03-14
Fix KeyboardScreenTests


CananArikan:
KeyboardScreenTests need to be fixed.

This issue is a sub-task of #405

PR #478, opened 2022-03-17
#114 Add support for resetting app data via launch argument

Clstroud:
Completes part of #114

# Description of Work
* Add ability to reset the app on launch via launch argument

# Notes to Test 
Screen Shot 2022-03-17 at 3 01 43 PM

And via XCUITest:
`XCUIApplication().launchArguments.append("resetAppDataOnLaunch")`

---


Clstroud:
Per @rhsalas3 on Slack -- we'll do the user-facing portion of this as a separate PR against the same ticket

rhsalas3:
Tested and confirmed the expected behavior for the launch argument for my tests (#464) and for @CananArikan's that are in review (#466 ).

This works great and eliminates the need for us to randomize our test data. We have the option of maintaining our custom categories/phrases still. We just need to be mindful of our base tests' setup() methods, which is where we set the launch argument.

CC: @rowildcard

rhsalas3 (approved):

PR #479, opened 2022-03-17
Fix Keyboard Screen Tests

### LINKED ISSUE #466: Fix KeyboardScreenTests

CananArikan:
closes #466 
The parent issue to https://github.com/willowtreeapps/vocable-ios/issues/466 is https://github.com/willowtreeapps/vocable-ios/issues/405

# Description of Work
Fixed broken keyboard tests. This is part of the effort to unlink the chained tests.

### DIFF
--- a/VocableUITests/Screens/MainScreen.swift
+++ b/VocableUITests/Screens/MainScreen.swift
@@ -51,4 +51,39 @@ class MainScreen: BaseScreen {
         let currentCategory = defaultCategories.count - (numTimesToScroll % defaultCategories.count)
         app.collectionViews.staticTexts[defaultCategories[currentCategory]].tap()
     }
+    
+    func scrollToMySayings() {
+        let mySayingsIdentifier = "category_title_cell_preset_user_favorites"
+        scrollRightFindCategoryAndTap(identifier: mySayingsIdentifier)
+    }
+    
+    private func scrollRightFindCategoryAndTap(identifier: String){
+       
+        // Loop through each category to find our category
+        let maxNumOfClickToReachCategory = 15
+        for _ in 1...maxNumOfClickToReachCategory {
+            if app.cells[identifier].exists {
+                app.cells[identifier].tap()
+                paginationRightButton.waitForExistence(timeout: 1)
+                break
+            } else {
+                categoryRightButton.tap()
+            }
+        }
+    }
+    
+    func findPhrase(phrase: String) {
+        
+        let predicate = NSPredicate(format: "label CONTAINS %@", phrase)
+        
+        // Loop through each custom category page to find our phrase
+        for _ in 1...totalPageCount {
+            if app.cells.staticTexts.containing(predicate).element.exists {
+                break
+            } else {
+                paginationRightButton.tap()
+            }
+        }
+    }
+    
 }
--- a/VocableUITests/Screens/MainScreen.swift
+++ b/VocableUITests/Screens/MainScreen.swift
@@ -27,18 +27,15 @@ class MainScreen: BaseScreen {
     
     // Find the current selected category and return it as a CategoryTitleCellIdentifier
     var selectedCategoryCell: CategoryTitleCellIdentifier {
-        let identifierPrefix = CategoryTitleCellIdentifier().categoryTitleCellPrefix
+        let identifierPrefix = CategoryTitleCellIdentifier.categoryTitleCellPrefix
         let identifierPredicate = NSPredicate(format: "identifier CONTAINS %@", identifierPrefix)
         let isSelectedPredicate = NSPredicate(format: "isSelected == true")
         
         // Build our query that first finds all category title cells, then finds among those the one that is selected
         let query = XCUIApplication().cells.containing(identifierPredicate).containing(isSelectedPredicate)
         
-        // Remove the identifierPrefix
-        let categoryTitleCellIdentifierString = query.element.identifier.replacingOccurrences(of: identifierPrefix, with: "")
-          
         // Return the selected category's full identifier
-        return CategoryTitleCellIdentifier(CategoryIdentifier(categoryTitleCellIdentifierString))
+        return CategoryTitleCellIdentifier(query.element.identifier)!
     }
     
     func isTextDisplayed(_ text: String) -> Bool {
--- a/VocableUITests/Screens/MainScreen.swift
+++ b/VocableUITests/Screens/MainScreen.swift
@@ -25,6 +25,22 @@ class MainScreen: BaseScreen {
     let paginationLeftButton = XCUIApplication().buttons["bottomPagination.left_chevron"]
     let paginationRightButton = XCUIApplication().buttons["bottomPagination.right_chevron"]
     
+    // Find the current selected category and return it as a CategoryTitleCellIdentifier
+    var selectedCategoryCell: CategoryTitleCellIdentifier {
+        let identifierPrefix = CategoryTitleCellIdentifier().categoryTitleCellPrefix
+        let identifierPredicate = NSPredicate(format: "identifier CONTAINS %@", identifierPrefix)
+        let isSelectedPredicate = NSPredicate(format: "isSelected == true")
+        
+        // Build our query that first finds all category title cells, then finds among those the one that is selected
+        let query = XCUIApplication().cells.containing(identifierPredicate).containing(isSelectedPredicate)
+        
+        // Remove the identifierPrefix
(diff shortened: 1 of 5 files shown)


VocableUITests/Tests/KeyboardScreenTests.swift
```diff
@@ -3,30 +3,59 @@
...
+        keyboardScreen.typeText(testPhrase)
+        keyboardScreen.favoriteButton.tap()
+        keyboardScreen.dismissKeyboardButton.tap()
+        mainScreen.scrollToMySayings()
+         
+        XCTAssertFalse(mainScreen.isTextDisplayed(testPhrase), "Expected the phrase \(testPhrase) to be deleted from 'My Sayings'")
+    }
+    
+
+
```
Clstroud:
```suggestion
```

VocableUITests/Screens/MainScreen.swift
```diff
@@ -51,4 +51,39 @@ class MainScreen: BaseScreen {
...
+        
+        // Loop through each custom category page to find our phrase
+        for _ in 1...totalPageCount {
+            if app.cells.staticTexts.containing(predicate).element.exists {
+                break
+            } else {
+                paginationRightButton.tap()
+            }
+        }
+    }
```
Clstroud: I like the structure y'all have come up with here! I've got one overarching suggestion we can discuss, but I want to emphasize that this is just me thinking aloud 😄
 
I think you might be able to find some harmony with the `XCUITest` framework by separating the notion of _locating_ a particular `XCUIElement` from _interacting_ with that element. Currently there is an inconsistency where one function specifies that it will find a category and also tap, but the other finds a phrase and doesn't really let the caller know if it was successful or not 🤭 

So in practice I might suggest something like (I'm making this up on the fly):
`func locateCategoryCell(matching identifier: String) -> XCUIElement?` and
`func locatePhraseCell(withText text: String) -> XCUIElement?`
You would be able to very cleanly add an assertion that your element exists (is non-nil) and then tap it like you would any other element.

You could even compose the two with a new function:
`func locatePhraseCell(withText text: String, inCategoryMatchingIdentifier categoryID: String) -> XCUIElement?`
in order to cut down on future boilerplate.

VocableUITests/Screens/MainScreen.swift
```diff
@@ -51,4 +51,39 @@ class MainScreen: BaseScreen {
         let currentCategory = defaultCategories.count - (numTimesToScroll % defaultCategories.count)
         app.collectionViews.staticTexts[defaultCategories[currentCategory]].tap()
     }
+    
+    func scrollToMySayings() {
+        let mySayingsIdentifier = "category_title_cell_preset_user_favorites"
```
Clstroud: It may make sense to have an enum of Categories where these identifiers can be stored. Then you could refer to those identifiers when scrolling to a particular category instead of having to make individual functions for each 🤔

rhsalas3: That's something I've been thinking about as well, while we start adding more and more identifiers. And especially now that we know that categories already have unique IDs we can reference.

To further clarify your suggestion when it comes to this particular function, does that look something like `func scrollToCategory(category: Category)` thus a call to it would be `mainScreen.scrollToCategory(Category.MySayings)`?

VocableUITests/Tests/KeyboardScreenTests.swift
```diff
@@ -3,30 +3,59 @@
...
         keyboardScreen.favoriteButton.tap()
         keyboardScreen.dismissKeyboardButton.tap()
-        mainScreen.scrollLeftAndTapCurrentCategory(numTimesToScroll: 1)
-
-        XCTAssert(mainScreen.isTextDisplayed(testPhrase), "Expected the phrase \(testPhrase) to be added to an displayed in 'My Sayings'")
+        
+        mainScreen.scrollToMySayings()
+        mainScreen.findPhrase(phrase: testPhrase)
+        
+        XCTAssertTrue(mainScreen.isTextDisplayed(testPhrase), "Expected the phrase \(testPhrase) to be added to an displayed in 'My Sayings'")
```
rowildcard: Typo. I believe you wanted your error message to say, "... added to _and_ displayed..."

VocableUITests/Tests/KeyboardScreenTests.swift
```diff
@@ -3,30 +3,59 @@
...
+       
+        mainScreen.keyboardNavButton.tap()
+        keyboardScreen.typeText(testPhrase)
+        keyboardScreen.favoriteButton.tap()
+        keyboardScreen.dismissKeyboardButton.tap()
+       
+        mainScreen.scrollToMySayings()
+        mainScreen.findPhrase(phrase: testPhrase)
+       
+        XCTAssertTrue(mainScreen.isTextDisplayed(testPhrase), "Expected the phrase \(testPhrase) to be added to an displayed in 'My Sayings'")
```
rowildcard: Ditto.
Typo. I believe you wanted your error message to say, "... added to _and_ displayed..."

VocableUITests/Tests/KeyboardScreenTests.swift
CananArikan: Fixed.

VocableUITests/Screens/MainScreen.swift
CananArikan: Functions are updated as suggested.

VocableUITests/Screens/MainScreen.swift
CananArikan: Added CategoryIdentifier/CategoryTitleCellIdentifier structs so that we now have a type to be used for passing along the category title cell element identifiers.

VocableUITests/Tests/CategoryIdentifier.swift
```diff
@@ -0,0 +1,48 @@
...
+
+struct CategoryTitleCellIdentifier {
+
+    private let categoryIdentifier: CategoryIdentifier
+    let categoryTitleCellPrefix: String = "category_title_cell_"
+    
+    // Empty: used to retrieve the prefix only
+    init() {
+        self.categoryIdentifier = CategoryIdentifier("")
+    }
```
Clstroud: I think you might be able to declare the prefix as `static let prefix: String = "category_title_cell_"`, which would mean you don't have to actually instantiate this struct to access it: `CategoryTitleCellIdentifier.prefix`

VocableUITests/Screens/MainScreen.swift
```diff
@@ -25,6 +25,22 @@ class MainScreen: BaseScreen {
...
+        let isSelectedPredicate = NSPredicate(format: "isSelected == true")
+        
+        // Build our query that first finds all category title cells, then finds among those the one that is selected
+        let query = XCUIApplication().cells.containing(identifierPredicate).containing(isSelectedPredicate)
+        
+        // Remove the identifierPrefix
+        let categoryTitleCellIdentifierString = query.element.identifier.replacingOccurrences(of: identifierPrefix, with: "")
+          
+        // Return the selected category's full identifier
+        return CategoryTitleCellIdentifier(CategoryIdentifier(categoryTitleCellIdentifierString))
```
Clstroud: I think it might make sense to move this logic into `CategoryTitleCellIdentifier` itself, giving it a [failable initializer](https://www.hackingwithswift.com/sixty/10/9/failable-initializers) that takes a `String`, and attempts to locate and remove the prefix.

Clstroud (commented):
I have just two small suggestions and then I think this will be 💯 Y'all are doing great

VocableUITests/Tests/CategoryIdentifier.swift
CananArikan: Fixed.

VocableUITests/Screens/MainScreen.swift
CananArikan: Updated as suggested.

Clstroud (approved):
Nicely done! 🚢 💯

rowildcard (approved):

VocableUITests/Tests/KeyboardScreenTests.swift
CananArikan: Fixed.

VocableUITests/Tests/KeyboardScreenTests.swift
CananArikan: Fixed.

PR #483, opened 2022-03-21
[114] Reset app settings via UI

### LINKED ISSUE #114: Reset App Settings

Clstroud:
Closes #114 

# Description of Work
* Add support for resetting the app via Vocable's Settings menu
* Add support for `.destructive` style in gazeable alert controller
* Shifts Core Data localization migration code out of the App Delegate and into its own entity
    - Needed so it could be called from more places (app launch + settings UI)
    - Sets it up for unit testing

## Notes to Test
No callouts as of yet


rowildcard:
Test verified on iPad (9th gen), iOS 15.2
☑︎ Delete a preset category - 123
☑︎ Delete a preset phrase - Basic need - I am xxx
☑︎ Rename preset category - General to Common
☑︎ Add a preset phrase - Basic need - Yo yo
☑︎ Add category & phrase - Pets - come Joe Joe
☑︎ Add favorite - Mom
☑︎ Speak a phrase to populate Recents 
☑︎ Reorder

Vocable/AppDelegate.swift
```diff
@@ -115,36 +115,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-            try deleteLegacyUserFavoritesCategoryIfNeeded(in: context)
-            try Category.updateAllOrdinalValues(in: context)
-
-            try context.save()
-        } catch {
-            assertionFailure(error.localizedDescription)
-        }
+    private func performPersistenceMigrationForCurrentLanguage() {
+        let migrationController = PersistenceMigrationController()
+        migrationController.performMigrationForCurrentLanguagePreferences()
```
moyerr, line 123: S 👏 R 👏 P 👏

I love seeing this stuff abstracted out into a dedicated controller 💯

Vocable/Common/Views/GazeableAlertController/GazeableAlertViewController.swift
```diff
@@ -143,6 +140,17 @@ private final class GazeableAlertButton: GazeableButton {
...
         contentEdgeInsets = .init(top: 24, left: 24, bottom: 24, right: 24)
+
+        setFillColor(.alertBackgroundColor, for: .normal)
+        setFillColor(.primaryColor, for: .selected)
+
+        if case .destructive = style {
+            setTitleColor(.errorRed, for: .normal)
+        } else {
+            setTitleColor(.black, for: .normal)
+        }
```
moyerr, line 153: At this point, it looks like both this method and `style`'s `didSet` block perform UI updates that are dependent on the `style` and the `traitCollection`. I wonder if we could unify these updates into a single method?

Vocable/CoreData/AppResetController.swift
```diff
@@ -32,7 +32,13 @@ struct AppResetController {
         guard resetPersistentStore() else {
             return false
         }
-        
+
+        let migrationController = PersistenceMigrationController()
```
moyerr: Since this `AppResetController` has its own instance of `NSPersistentContainer`, should we be passing that reference down into the `PersistenceMigrationController`?

Vocable/Extensions/UIColor+AppExtensions.swift
```diff
@@ -72,4 +72,5 @@ extension UIColor {
     static let alertBackgroundColor = UIColor(safelyNamed: "AlertBackground")!
 
     static let grayDivider = UIColor(safelyNamed: "GrayDivider")!
+    static let errorRed = UIColor(safelyNamed: "ErrorRed")!
```
moyerr, line 75: Could we drop the word "Red" from this to keep the name purely semantic? If `error` is too general of a name, maybe we could do something like `primaryError` or `errorText`

Vocable/Common/Views/GazeableAlertController/GazeableAlertViewController.swift
```diff
@@ -13,6 +13,7 @@ final class GazeableAlertAction: NSObject {
     public enum Style {
```
moyerr: I know this isn't new code, but it seems like as good a time as any to remove `public` here.

moyerr (commented):
Good stuff! Big fan of the abstractions and the `destructive` button style. Just had a bit of feedback, some of which we've already discussed, and most of which is fairly minor 😃

Vocable/CoreData/AppResetController.swift
Clstroud: Whoops! Thanks, this is actually a great catch.

Vocable/Extensions/UIColor+AppExtensions.swift
Clstroud: We probably can, and I agree with this sentiment, but this is the name that exists in the Figma style guide so I'm not sure if we should deviate?

moyerr: Got it, totally understand. Yeah, I'm down with keeping parity between our color names and the colors in Figma. I had not seen the style guide in figma yet, so I'll be checking that out today 👀

moyerr (approved):
🏅

rowildcard:
The app crashed on me yesterday after tapping reset. Sashank was able to reproduce the crash today so I'm going to work on narrowing this down. 
I have the crash log from the test device if needed.

rowildcard:
Verified the most recent commit to handle a crash. ✅ 

**STEPS**
1. On the Main screen tap the Keyboard button. 
2. Add a word. 
3. Repeat so there are three new words in "My Sayings".
4. Go to Settings and tap the Reset App Settings button. 
5. On the confirmation pop-up select "Reset".

moyerr (approved):
Nice work on the updates 💪

Issue #481, opened 2022-03-18
Vocable iOS - Empty State text mismatch


SashankPatel:
**Describe the bug**
There is a mismatch in the empty state text between the code and figma.
App: "This category has no phrases yet"
Figma: "You don't have any phrases saved yet"

After chatting with a designer, it is preferred to have the figma text rather than the text in the app.

**To Reproduce**

1. Launch the app
2. Navigate to Settings > Categories and Phrases.
3. Tap on the + button to add a category.
4. Proceed with steps to adding a new category.
5. After Category is added, go into the newly created category and tap on the "Edit Phrases" button

**Expected behavior**
Empty state to appear with the following design:
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=972%3A8837

**Actual behavior**
Empty state appears as shown in figma but with the text "This category has no phrases yet"

**Device Information**
This is device independent

SashankPatel:
Crowdin was updated with the new string.  We are looking into how crowdin distributes those changes, i.e. will a PR be automatically generated?

SashankPatel:
CC @moyerr Assigned this ticket to you as well, initially we didn't think there was any work left for #253.

Issue #463, opened 2022-03-11
Vocable iOS - Custom Category tile highlighting on Home screen.


SashankPatel:
**Device Information**
Seen on an iPad 7th Gen with iOS 14.

**Describe the bug**
On the Home screen in the area where you can select categories, there is some inconsistent highlighting.

**To Reproduce**
This was seen on the a
1. Launch the app
2. Create a custom category if there arn't any.
3. Navigate back to the home screen.
4. Tap on the Listen category -> Category tile should be highlighted.
5. Tap on the new Custom Category next to the Listen <- ISSUE 1
6. Tap on the add phrase button and add a new phrase. <- ISSUE 2

**Expected behavior**
Listen should unhighlight and the custom category tile should highlight.
After a phrase is added from the main screen the custom category and phrase should be visible.

**Actual behavior**
Both the Listen and Custom Category tiles are highlighted.
This also happens with General, if you tap General and then the Custom category both will be highlighted.
After a phrase is added, instead of showing the custom category and newly added phrase, the Listen/General tab is displayed.

**Screenshots**
![IMG_0032](https://user-images.githubusercontent.com/50410188/157957313-1847fab7-6aca-40c9-817d-9516679c2966.PNG)

rhsalas3:
Per my update to [the comment](https://github.com/willowtreeapps/vocable-ios/pull/491#issuecomment-1076243145) left on PR #491 (to implement feature #253 ):

Here is the recording of the additional behavior noted, whereby the custom category is not automatically selected as a user scrolls to it on an iPhone in Portrait. Also in this recording, while in Landscape the behavior captured in this bug issue is demonstrated.

Simulator - iPhone 12 - iOS 15.4 - from branch `feature/253/categories-sayings-empty-state`

https://user-images.githubusercontent.com/34660497/159769895-a48d9c47-de73-449b-9900-aa67980d0771.mp4

SashankPatel:
Verified on iOS 14 and 15, this but seems to no longer occur on 1.4.0-2485.  Verified on iPhone 12 and iPad 7th gen.  Closing out the bug.

PR #491, opened 2022-03-22
#253 | Enable and update Empty State views for Sayings & Categories

### LINKED ISSUE #253: Empty states for My Sayings and Custom Categories
**As a vocable app user, I would like to see an empty state message when I have just added a category**

NOTE: As of 3/21/22 this is mostly implemented.  It is hidden via a feature flag therefore wont appear in the App store and Adhoc schemes.  There may be a need for some design updates, but when that is worked on as part of this ticket, we will need to remove the feature flag.

Acceptance Criteria
- After successfully adding a category, navigating to the category on the home screen display an empty state screen.
- After successfully adding a category, navigating to the detail view of that category in Settings would display an empty state screen.
- If there are no `My Sayings` an empty state screen will be displayed. 

Design on Home Screen:
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=1%3A141
Design in Settings:
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=1%3A1344

moyerr:
### Closes #253 

--------
# Description of Work

This PR had two main goals:

1. Remove the feature flag that was preventing the Empty State from appearing on non-Development builds.
2. Update the styling of the `EmptyStateView` to better reflect what is captured in the designs.

## Screenshots

| Orientation | iPhone 13 | iPad Mini |
| :--- | :---: | :---: |
| Portrait |  |  |
| Landscape |  |  |


Vocable/Common/Views/EmptyStateView.swift
```diff
@@ -192,40 +203,29 @@ final class EmptyStateView: UIView {
...
-        }
-
-        if descriptionLabel.attributedText != nil {
-            stackView.insertSubview(descriptionLabel, belowSubview: titleLabel)
-        } else {
-            stackView.removeArrangedSubview(descriptionLabel)
-            descriptionLabel.removeFromSuperview()
-        }
+        imageView.isHidden = imageView.image == nil
+        descriptionLabel.isHidden = descriptionLabel.text == nil && descriptionLabel.attributedText == nil
```
Clstroud, line 229: Ha! Yes, thank you. Good catch.

Vocable/Extensions/UIEdgeInsets+Operators.swift
```diff
@@ -27,4 +27,17 @@ extension UIEdgeInsets {
...
+        .init(top: inset, left: inset, bottom: inset, right: inset)
+    }
+
+    static func vertical(_ verticalInset: CGFloat) -> UIEdgeInsets {
+        .init(top: verticalInset, left: 0, bottom: verticalInset, right: 0)
+    }
+
+    static func horizontal(_ horizontalInset: CGFloat) -> UIEdgeInsets {
+        .init(top: 0, left: horizontalInset, bottom: 0, right: horizontalInset)
+    }
```
Clstroud, line 41: Probably not a blocker, but something to consider: `horizontal` and `vertical` threw me for a loop up above because I couldn't tell if the inset was being uniformly applied to each edge, or if the total inset was being divided among each edge along that axis. APIs [like this one](https://developer.apple.com/documentation/coregraphics/cgrect/1454218-insetby) on CGRect are probably what make me overthink things like this, but alas.

Having them be static functions is not bad, but I'm almost wondering if we would get more call site clarity out of following the SwiftUI-ish pattern of having `UIEdgeInsets(top: CGFloat = .zero, left: CGFloat = .zero, bottom: CGFloat = .zero, right: CGFloat = .zero)`?

Also: I know they're existing code but the operators _also_ threw me off, so maybe I'm just having a case of EOD.

Vocable/Extensions/StringProtocol+Range.swift
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /// Converts a range of Indices to its `NSRange` representation.
+    ///
+    /// - Parameter rangeExpression: The range of string indices
+    /// - Returns: The `NSRange` representation of the `rangeExpression`
+    func nsRange<R: RangeExpression>(from rangeExpression: R) -> NSRange where R.Bound == Index {
+        NSRange(rangeExpression, in: self)
+    }
+}
```
Clstroud: I'm definitely into the idea of having the range of indices available in a more convenient way, but if we're extending something as generalized as `StringProtocol` I would love to see something that feels more like it's part of Foundation if that makes sense? Something like `indexRange` or `rangeOfIndices`? Just typing out loud here 🤔 

I'm not personally in love with `StringProtocol` being able to produce `NSRange` values, in part because it feels like we're overloading that type unnecessarily when `NSRange` gives us a really clear initializer... but also `ns` being lowercased just feels wrong. Especially when you consider line 19 where it reads like an `NSRange` initializer but in SpONgeBob case 😛 

I know we want to shift more in the direction of attributed strings for text styling going forward, and I wonder if there's anything we can do to extend attributed string to be more amenable to Swift strings as opposed to shoving that responsibility downward into `StringProtocol` 🤔 A discussion for another day!

Clstroud (approved):
Leaving a few notes to chat about, but this is a solid first PR! Leavings things tidier than you found them and that's awesome 🎉

SashankPatel:
Test Notes 2/23:
-Verified on Iphone 14.4 and 15.4 and iPad 15.4

- Verified empty state on main screen for My Sayings
- Verified layout with different orientations.
- Verified empty state in My Saying Settings category with various orientations.
- Verified empty state with new custom category in settings and main screen.
- Verified empty state is replaced by phrase, when adding via main screen or settings.
- Verified empty state is visible when deleting the only phrase in the category.
- Verified feature flag is removed by running the adhoc build.
- Noticed a bug related to #463, when a phrase is added but that will be captured in a separate issue.
    - [from Rudy Salas - 3/23/22] Expanding on this bug related to 463: on this PR branch I am seeing the captured behavior but also that a new category does not automatically get selected when scrolling to it on an iPhone (via simulator). I will add a recording of this behavior to the bug ticket.
- Empty state text does not match designs, but that will be covered in bug ticket: #481

SashankPatel (approved):
Good to go from Test.

rhsalas3:
@moyerr  and @sgordon1024 

Are we removing the pagination from the screen when it's in an empty state? The [Figma designs](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=1%3A141) show the pagination elements. I have not looked at the work on this PR branch, I am just going off of the screenshots shared in the description of this PR.

I just want to make sure because I was actually just working on the existing pagination tests, and will make adjustments for this.

***Update***:
[Left the same comment in Figma and Steve is content to have the buttons removed](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR?node-id=157:471#169317172). He has since updated the designs in Figma to reflect this.

CC: @SashankPatel @rowildcard @CananArikan

SashankPatel:
This is still looking good after the latest commit: :ship:

Noticed an UI issue with orientation but that is not related and will be captured as a separate bug.

PR #496, opened 2022-03-23
#476 | Custom Category Refactor

### LINKED ISSUE #476: Update Layout in the Phrases view based on designs
**Acceptance Criteria**:
No Functional changes.
- As a user I shall be able to delete phrases.
- As a user I shall be able to edit phrases
- As a user I shall be able to add new phrases.
- As a user I shall be able to scroll through multiple pages of phrases

The layout on the Edit Phrase view has updated.  The delete confirmation button should be red. This ticket is to make updates to the existing UI to implemented the design updates.

**Design**:
iPad Landscape: https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=969%3A1629
iPad Portrait: https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=3398%3A19036
iPhone: https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=3356%3A17664

jsmorgan42:
closes #476 

# Description of Work
This PR refactors the collection view cells in the custom category edit phrase screen so that there are larger separated buttons that are tappable rather than embedding smaller buttons into the overall collection view cell. This makes the actions more easily accessible for the user.

It also builds the groundwork to rely on the iOS 14 feature using configurable cells. This will allow us to reduce the amount of extraneous cell declarations and rather rely on the cell configuration to create a composable and reusable cell type. In future work, we will want to most likely want to figure out how to link the "primary action" to the cell selection so that we can better leverage the collection view delegate functions.

## Notes to Test (Optional)

- This PR increases the minimum iOS version target to iOS 14.
- There is a known bug where the layout change animation will make the data disappear for a brief second on iOS 15.

---


rhsalas3:
@sgordon1024 @SashankPatel @jsmorgan42 @Clstroud 

I have tested the new layout on this PR branch. Everything looks correct except for one potential concern. On an iPhone in landscape, there is only one column of phrases per page but the designs in Figma suggest there should be 2 columns. I left a comment on the designs to see what the team's preference is. I will wait for an answer before making this as "verified".

Design
https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=3428%3A24868

Screenshot from branch:
EditPhrasesScreen_iPhone12_landscape

---
***Update***
Spoke to Chris and Sashank in a Slack huddle. We think it's best to iterate on it if we decide to make any changes for the iPhone landscape layout to match the design. Once Steve gets a chance to take a look at it, and depending on what he says, we'll just add a new ticket for any necessary changes.

rhsalas3:
Testing complete. Looks good to me.

Vocable/Common/CarouselGridLayout.swift
```diff
@@ -360,7 +375,7 @@ final class CarouselGridLayout: UICollectionViewLayout {
 
     override func targetContentOffset(forProposedContentOffset proposedContentOffset: CGPoint, withScrollingVelocity velocity: CGPoint) -> CGPoint {
         guard let collectionView = collectionView, pagesPerSection > 0,
-            pageContentSize.width > 0 || interItemSpacing > 0 else { return proposedContentOffset }
+              pageContentSize.width > 0 || interItemSpacing > .zero else { return proposedContentOffset }
```
Clstroud:
```suggestion
              pageContentSize.width > .zero || interItemSpacing > .zero else { return proposedContentOffset }
```

Vocable/Common/CarouselGridLayout.swift
```diff
@@ -90,7 +90,23 @@ final class CarouselGridLayout: UICollectionViewLayout {
         }
     }
 
-    var interItemSpacing: CGFloat = 0 {
+    struct InterItemSpacing: Equatable {
+        let interRowSpacing: CGFloat
+        let interColumnSpacing: CGFloat
+
+        static let zero: Self = .init(interRowSpacing: 0, interColumnSpacing: 0)
```
Clstroud:
```suggestion
        static let zero: Self = .init(interRowSpacing: .zero, interColumnSpacing: .zero)
```

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+    // MARK: - Subviews
+
+    private lazy var stackView: UIStackView = {
+        let stackView = UIStackView()
+        stackView.axis = .horizontal
+        stackView.alignment = .center
+        stackView.distribution = .fill
+        stackView.spacing = 8
+        return stackView
+    }()
```
Clstroud: Note for future: I could definitely imagine having a `UIStackView` initializer with nil default values that could be overridden to express this more concisely. Not sure if it's a huge ergonomic improvement, but it would at least cut down on the need for some of these closures.

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+        updateLeadingActionAccessoryButtons(with: configuration)
+        updatePrimaryLabelButton(with: configuration)
+    }
+
+    private func updatePrimaryLabelButton(with configuration: VocableListContentConfiguration?) {
+        primaryLabelButton.contentHorizontalAlignment = .left
+        primaryLabelButton.setAttributedTitle(configuration?.attributedTitle, for: .normal)
+        primaryLabelButton.addTarget(self, action: #selector(handlePrimaryActionSelection(_:)), for: .primaryActionTriggered)
+
+        let trailingInsets: NSDirectionalEdgeInsets = .init(top: 0, leading: 16, bottom: 0, trailing: 16)
```
Clstroud:
```suggestion
        let trailingInsets = NSDirectionalEdgeInsets(top: .zero, leading: 16, bottom: .zero, trailing: 16)
```

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+            primaryLabelButton.setTrailingAccessoryView(nil, insets: .zero)
+        }
+    }
+
+    private func updateLeadingActionAccessoryButtons(with configuration: VocableListContentConfiguration?) {
+
+        let actions = configuration?.actions ?? []
+
+        // Ensure the minimum number of action buttons are present
+        let numberOfButtonsNeeded = max(actions.count - accessoryButtonStackView.arrangedSubviews.count, 0)
```
Clstroud:
```suggestion
        let numberOfButtonsNeeded = max(actions.count - accessoryButtonStackView.arrangedSubviews.count, .zero)
```

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+        (0 ..< numberOfButtonsNeeded).forEach { _ in
```
Clstroud:
```suggestion
        (.zero ..< numberOfButtonsNeeded).forEach { _ in
```

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+            insertAccessoryButton()
+        }
+
+        // Update existing buttons to match new states
+        let arrangedButtons = accessoryButtonStackView.arrangedSubviews.compactMap { $0 as? GazeableButton }
```
Clstroud, line 103: I wish [this](https://github.com/apple/swift-algorithms/blob/main/Guides/Compacted.md) were in the standard lib, but I could also imagine an extension allowing e.g. `compacted(as: GazeableButton.self)` that _might_ just be idiomatic and ergonomic enough to be worthwhile.

Clstroud (commented):
I was pretty involved with this, so I'm just going to submit a couple comments/observations. The few things that stand out to me were already known and discussed with the team.

Vocable.xcodeproj/project.pbxproj
```diff
@@ -1439,7 +1471,7 @@
 				);
 				HEADER_SEARCH_PATHS = "";
 				INFOPLIST_FILE = "Vocable/Supporting Files/Info.plist";
-				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
```
moyerr, line 1478: 👀

Vocable/Common/CarouselGridLayout.swift
```diff
@@ -90,7 +90,23 @@ final class CarouselGridLayout: UICollectionViewLayout {
...
+        static func uniform(_ spacing: CGFloat) -> Self { .init(interRowSpacing: spacing, interColumnSpacing: spacing) }
+
+        static func > (lhs: InterItemSpacing, rhs: InterItemSpacing) -> Bool {
+            return lhs.interRowSpacing + lhs.interColumnSpacing > rhs.interRowSpacing + rhs.interColumnSpacing
+        }
+
+        static func < (lhs: InterItemSpacing, rhs: InterItemSpacing) -> Bool {
+            return lhs.interRowSpacing + lhs.interColumnSpacing < rhs.interRowSpacing + rhs.interColumnSpacing
+        }
```
moyerr: A couple thoughts here:

1. To support greater-than/less-than comparisons, it is recommended to conform to the `Comparable` protocol, which will actually only require you to implement one of these two operators.
2. I don't think it's clear what it actually means to compare two `InterItemSpacing` values. I find it interesting you chose to compare the sum of the two values - what does the sum represent? I have some alternatives in mind, but before I go down that rabbit hole, is the comparison something we actually need? Would it not be sufficient to check if the spacing is non-zero?

Vocable/Common/NSLayoutConstraint+Priority.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+
+import UIKit
+
+extension NSLayoutConstraint {
+
+    func withPriority(_ priority: UILayoutPriority) -> NSLayoutConstraint {
+        self.priority = priority
+        return self
+    }
```
moyerr, line 16: Good shit 💯 👌 👀

Vocable/Common/Views/GazeableButton.swift
```diff
@@ -148,6 +159,45 @@ class GazeableButton: UIButton {
...
+    func setTrailingAccessoryView(_ view: UIView?, insets: NSDirectionalEdgeInsets) {
+
+        defer {
+            trailingAccessoryView = view
+        }
+
+        if let trailingAccessoryView = trailingAccessoryView {
+            if view == trailingAccessoryView {
+                return
+            }
```
moyerr: Would it make more sense to use the identity operator (`===`) here?

Vocable/Common/Views/GazeableButton.swift
```diff
@@ -148,6 +159,45 @@ class GazeableButton: UIButton {
...
+        ])
+    }
+
+    override func layoutSubviews() {
+
+        let layoutGuideWidth = trailingAccessoryViewLayoutGuide.layoutFrame.width
+
+        if self.contentEdgeInsets.right != layoutGuideWidth {
+            self.contentEdgeInsets.right = layoutGuideWidth
+        }
```
moyerr: What if `contentEdgeInsets.right` is greater than the `layoutGuideWidth`? In that case the trailing accessory view would have enough room, so should we still be messing with the `contentEdgeInsets` in that case?

Vocable/Common/VocableListCell/VocableListCellAccessory.swift
```diff
@@ -0,0 +1,53 @@
...
+    let content: Content
+    let isEnabled: Bool
+
+    private static func systemImage(_ imageName: String, symbolConfiguration: UIImage.SymbolConfiguration? = nil) -> UIImage {
+        var image = UIImage(systemName: imageName)!
+        if let symbolConfiguration = symbolConfiguration {
+            image = image.applyingSymbolConfiguration(symbolConfiguration)!
+        }
+        return image
+    }
```
moyerr: This functionality seems like it could be useful as an extension on `UIImage`

Vocable/Common/VocableListCell/VocableListCellAccessory.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    static func toggle(isOn: Bool, isEnabled: Bool = true) -> VocableListCellAccessory {
+        return VocableListCellAccessory(content: .toggle(isOn: isOn), isEnabled: isEnabled)
+    }
+
+    static func == (lhs: VocableListCellAccessory, rhs: VocableListCellAccessory) -> Bool {
+        lhs.isEnabled == rhs.isEnabled &&
+        lhs.content == rhs.content
+    }
```
moyerr: Can this not be synthesized?

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+    private func setupViews() {
+        backgroundColor = nil
+
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(stackView)
+
+        accessoryButtonStackView.isHidden = true
+        stackView.addArrangedSubview(accessoryButtonStackView)
+
+        stackView.addArrangedSubview(primaryLabelButton)
```
moyerr: Super minor nitpick, and not something that needs to be changed by any means, but almost all of these can go in the closures where the stack views are configured. I think the only thing I'd leave here is the `addSubview(stackView)`

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+            primaryLabelButton.heightAnchor.constraint(equalTo: stackView.heightAnchor)
+        ])
+
+        configure(with: configuration)
+    }
+
+    private func configure(with configuration: UIContentConfiguration) {
+        guard let configuration = configuration as? VocableListContentConfiguration else {
+            updateLeadingActionAccessoryButtons(with: nil)
+            updatePrimaryLabelButton(with: nil)
```
moyerr: I don't think this is wrong at all, just wondering if it might just make more sense to completely ignore any invalid configuration types rather than wiping out the values we have? Apple didn't make this situation easy by forcing us to deal with `any UIContentConfiguration` 😅

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+            return
+        }
+
+        updateLeadingActionAccessoryButtons(with: configuration)
+        updatePrimaryLabelButton(with: configuration)
+    }
+
+    private func updatePrimaryLabelButton(with configuration: VocableListContentConfiguration?) {
+        primaryLabelButton.contentHorizontalAlignment = .left
```
moyerr: Is this something that should live in the `configuration`?

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+        primaryLabelButton.setAttributedTitle(configuration?.attributedTitle, for: .normal)
+        primaryLabelButton.addTarget(self, action: #selector(handlePrimaryActionSelection(_:)), for: .primaryActionTriggered)
```
moyerr, line 89: Should we be removing the old target-action first?

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+            }
+        case .toggle(let isOn):
+            if let trailingToggle = primaryLabelButton.trailingAccessoryView as? UISwitch {
+                trailingToggle.setOn(isOn, animated: true)
+            } else {
+                let toggle = UISwitch()
+                toggle.setOn(isOn, animated: true)
+                toggle.isUserInteractionEnabled = false
+                primaryLabelButton.setTrailingAccessoryView(toggle, insets: trailingInsets)
+            }
```
moyerr: I wonder if we could move the accessory functionality out of `GazeableButton` and into a specialized sublcass that supports a limited number of accessory types. That way, we can just tell the view to configure itself for a given accessory type, rather than instantiating new views inline during the configuration here. 

Then, as an implementation detail, that subclass could initialize those accessory types lazily, so that if they are never needed during the lifetime of the view, they will never take up any memory. However, on the flipside, if we have a situation where we flip back and forth between accessory types multiple times, we would only ever need to initialize them once, rather than throwing them away each time the accessory changes.

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+
+        // If there are no buttons to display, hide their stack view
+        accessoryButtonStackView.isHidden = arrangedButtons.allSatisfy(\.isHidden)
+    }
+
+    private func insertAccessoryButton() {
+        let button = GazeableButton()
+        accessoryButtonStackView.addArrangedSubview(button)
+        NSLayoutConstraint.activate([
+            button.heightAnchor.constraint(equalTo: button.widthAnchor, multiplier: 1),
```
moyerr, line 130: `button.automaticallyTranslates..... = false`?

Vocable/Common/VocableListCell/VocableListCellContentView.swift
```diff
@@ -0,0 +1,187 @@
...
+
+        guard let buttonIndex = accessoryButtonStackView.arrangedSubviews.firstIndex(of: sender) else {
+            return
+        }
+        guard let configuration = configuration as? VocableListContentConfiguration else {
+            return
+        }
+        guard let accessory = configuration.actions[safe: buttonIndex] else {
+            return
+        }
```
moyerr: Could we collapse these into a single `guard`?

moyerr (commented):
I'm really starting to like this cell configuration stuff! Left some minor feedback and some ideas we can talk through on Monday!

Vocable/Common/Views/GazeableButton.swift
Clstroud: Good catch, yes it would (though I believe that's the default for UIViews because of NSObject?). Would be clearer about intent either way!

Vocable/Common/Views/GazeableButton.swift
Clstroud: Just to slightly clarify: the goal here is to cautiously avoid a recursive layout pass by ensuring we aren't setting `contentEdgeInsets` to a redundant value. 

If I'm hearing you correctly and the layout guide were to resolve to something narrower than its previous configuration, we would have a wider gap between the content of the button and the trailing view than was prescribed by the caller, which we wouldn't want. Is that right?

Vocable/Common/VocableListCell/VocableListCellAccessory.swift
Clstroud: It can probably be synthesized for `VocableListCellAccessory`, but I think this implementation may need to have `Equatable` defined explicitly for `Content` because of `UIImage`. Same reason earlier versions of this type had to manually define it here.

Vocable/Common/VocableListCell/VocableListCellContentView.swift
Clstroud: Not sure we have any cases where this wouldn't be the desired value, but it also looks like this gets defined in its initialization closure, so I don't think this line is actually needed?

Vocable/Common/VocableListCell/VocableListCellContentView.swift
Clstroud: Not necessary:

> You may call this method multiple times to configure multiple targets and actions for the control. It is also safe to call this method multiple times with the same values for the target and action parameters. The control maintains a list of its attached targets and actions along with the events each supports.

This was effectively why I went this route instead of `UIAction` 😄

Vocable/Common/VocableListCell/VocableListCellContentView.swift
Clstroud: `UIStackView` owns the button's layout, so it gets to decide. We're just piling on extra constraints.

Vocable/Common/CarouselGridLayout.swift
Clstroud: @moyerr This is a great callout that I totally missed but you're right.

I have a feeling this may need to half-revert back to having individual private properties in the layout that can be compared with mutual exclusion. Could still be configured/assigned this way, but applied to privately maintained `interRowSpacing` and `interColumnSpacing` properties.

Clstroud: Pure speculation, but I wonder if this could have anything to do with the wacky scrolling we're seeing in basically all of the collection views? Not convinced that makes sense, just throwing it out there

Clstroud (approved):
Good work @jsmorgan42  😄

PR #498, opened 2022-03-24
Disable Bitcode

jsmorgan42:
# Description of Work
An invalid bundle error is currently being thrown when building on CI currently ([for context](https://developer.apple.com/forums/thread/701949)). We need to disable bitcode until the issue get fixed in a future version of Xcode. 

---


Clstroud (approved):
🤞🏻

PR #512, opened 2022-03-28
Re-enable CustomCategoriesTests - Refactor to leverage reset at launch

### LINKED ISSUE #511: [Test] Re-enable CustomCategoriesTests
We accidentally disabled the CustomCategoriesTests somehow. We should re-enable them so that we can get the test results during builds.

AC: 
All tests in CustomCategoriesTests are enabled and run during builds.

rhsalas3:
Closes https://github.com/willowtreeapps/vocable-ios/issues/511

# Description
At some point the CustomCategoriesTests.swift class became disabled for test runs. I have re-enabled the suite so that all of the tests we expect to run, will. 

I also took the time to refactor what we forgot to do earlier, around randomization. We no longer need to randomize the name of our categories because we implemented the reset feature w/in the app.

## Notes to Test
All tests should pass on iPhone and iPad.

### DIFF
--- a/VocableUITests/Tests/CustomCategoriesTests.swift
+++ b/VocableUITests/Tests/CustomCategoriesTests.swift
@@ -8,7 +8,7 @@
 
 import XCTest
 
-class CustomCategoriesTest: CustomCategoriesBaseTest {
+class CustomCategoriesTests: CustomCategoriesBaseTest {
 
     func testAddNewPhrase() {
         let customPhrase = "dd"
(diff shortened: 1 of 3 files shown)


VocableUITests/Tests/CustomCategoriesTests.swift
```diff
@@ -99,8 +96,9 @@ class CustomCategoriesTest: CustomCategoriesBaseTest {
...
     }
     
-    // TODO: Disabled for now. Moving it to a different test class as part of issue #405
-    // https://github.com/willowtreeapps/vocable-ios/issues/405
+    // TODO: Disabled for now. Moving it to a different test class as part of issue #405; tracked in issue #470
+    // https://github.com/willowtreeapps/vocable-ios/issues/470 -> Implementation
+    // https://github.com/willowtreeapps/vocable-ios/issues/405 -> Parent
     func testPagination() {
         let customCategoryThree = "Testc"
         let createdCustomCategory = ("9. "+customCategoryThree)
```
Clstroud, line 104: I think my branch is going to break this since the numbers are going away, but that's okay. Out of scope for this PR.

Clstroud (approved):
Short and sweet🥇 🚀

SashankPatel:
@rhsalas3 The pagination test is failing, which is expected, and being revamped in another PR.  Will enabled this block merging of PR's until the pagination tests are fixed?

VocableUITests/Tests/CustomCategoriesTests.swift
rhsalas3: Good call out. We've actually already refactored our other tests to not require the number. I think we'll be in good shape even with the changes in your branch! 💯

Issue #492, opened 2022-03-23
Update Category List UI


Ly-Kyli:
**Acceptance Criteria**:

1. The space for each category name now spans to fill the width of the screen (currently: the user has to land on the arrow on open each category)
2.  New design reflects less negative space and allows user more room to select each category

**Design**:

[Ipad Landscape](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=461%3A0)
[Ipad Portrait](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=461%3A901)
[Iphone Landscape](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=461%3A1214)
[Iphone Portrait](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=461%3A1293)

PR #520, opened 2022-03-28
Fixed Settings Screen Tests

### LINKED ISSUE #480: Fix SettingScreenTests
SettingScreenTests need to be fixed.

This issue is a sub-task of https://github.com/willowtreeapps/vocable-ios/issues/405

CananArikan:
testHideShowToggle is fixed, closes #480 
The parent issue to #480 is #405 

# Description of Work
Fixed broken settings screen test. This is part of the effort to unlink the chained tests.

### DIFF
--- a/VocableUITests/Screens/SettingsScreen.swift
+++ b/VocableUITests/Screens/SettingsScreen.swift
@@ -36,9 +36,10 @@ class SettingsScreen: BaseScreen {
     }
     
     func locateCategoryCell(_ category: String) -> XCUIElementQuery {
+        let predicate = NSPredicate(format: "label CONTAINS %@", category)
         // Loop through each page to find our category
         for _ in 1...totalPageCount {
-            if categoryCellQuery(category).element.exists{
+            if cells.staticTexts.containing(predicate).element.exists{
                 break
             } else {
                 settingsPageNextButton.tap()
--- a/VocableUITests/Screens/SettingsScreen.swift
+++ b/VocableUITests/Screens/SettingsScreen.swift
@@ -3,7 +3,8 @@
 //  VocableUITests
 //
 //  Created by Kevin Stechler on 5/19/20.
-//  Copyright © 2020 WillowTree. All rights reserved.
+//  Updated by Canan Arikan and Rudy Salas on 03/28/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
 //
 
 import XCTest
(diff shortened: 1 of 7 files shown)


VocableUITests/Tests/CustomCategoriesTests.swift
```diff
@@ -13,6 +13,7 @@ class CustomCategoriesTests: CustomCategoriesBaseTest {
     func testAddNewPhrase() {
         let customPhrase = "dd"
         let confirmationAlert = "Are you sure? Going back before saving will clear any edits made."
```
Clstroud: We should probably give this an identifier instead of keying off of the actual string, since this is subject to change and would break the test?

Clstroud: Looks like that identifier is currently `text_editor.alert.cancel_editing_confirmation.title`

VocableUITests/Tests/SettingsScreenTests.swift
```diff
@@ -3,45 +3,42 @@
...
+
+        // Verify that when the category is shown, up and down buttons are enabled.
+        settingsScreen.locateCategoryCell(category).buttons[settingsScreen.categoryForwardButton].tap()
+        settingsScreen.showCategorySwitch.tap()
+        settingsScreen.leaveCategoryDetailButton.tap()
+        
+        settingsScreen.locateCategoryCell(category)
+        XCTAssertTrue(settingsScreen.locateCategoryCell(category).buttons[settingsScreen.categoryUpButton].isEnabled)
+        XCTAssertTrue(settingsScreen.locateCategoryCell(category).buttons[settingsScreen.categoryDownButton].isEnabled)
+        XCTAssertTrue(settingsScreen.locateCategoryCell(category).buttons[settingsScreen.categoryForwardButton].isEnabled)
```
Clstroud:
```suggestion
        let categoryCell = settingsScreen.locateCategoryCell(category)
        guard categoryCell.waitForExistence(20) else 
            XCTFail("Couldn't locate category cell")
            return
        }        
        XCTAssertTrue(categoryCell.buttons[settingsScreen.categoryUpButton].isEnabled)
        XCTAssertTrue(categoryCell.buttons[settingsScreen.categoryDownButton].isEnabled)
        XCTAssertTrue(categoryCell.buttons[settingsScreen.categoryForwardButton].isEnabled)
```
Just using this one snippet as an example, I think you can grab the cell element once and store it as a variable to reference in the assertions below it. I also threw in a `waitForExistence` because I'm not sure exactly what happens if `locateCategoryCell()` is unsuccessful, but [there are other strategies](https://mokacoding.com/blog/write-better-swift-tests-with-xctest-assertions/) that might help with that. Not a huge deal, mostly just wanted to grab the cell once for performance and readability reasons 😄

VocableUITests/Screens/SettingsScreen.swift
```diff
@@ -17,39 +18,51 @@ class SettingsScreen: BaseScreen {
...
                 break
             } else {
                 settingsPageNextButton.tap()
             }
         }
+        
+        return categoryCellQuery(category)
+    }
+    
+    func categoryCellQuery(_ category: String) -> XCUIElementQuery {
```
Clstroud: Does this get used outside this context? I wonder if maybe it can be `private`?

VocableUITests/Tests/SettingsScreenTests.swift
```diff
@@ -54,18 +51,19 @@ class SettingsScreenTests: BaseTest {
         XCTAssert(settingsScreen.otherElements.containing(.staticText, identifier: generalCategoryText).element.exists)
         XCTAssert(settingsScreen.otherElements.containing(.staticText, identifier: basicNeedsCategoryText).element.exists)
```
Clstroud, line 52: Based on some of the test failures I've seen recently, we should probably be using `waitForExistence()` to handle slow transition animations. I'm worried this could potentially be flaky otherwise, but I might be missing some context.

Clstroud: Also: same note about storing elements as variables. `settingsScreen.otherElements.containing(.staticText, identifier: generalCategoryText).element` is _very_ long and difficult to read multiple times 😅

Clstroud (changes requested):
Structurally this seems pretty solid, but I left some feedback notes for tightening things up a bit. Happy to discuss any/all suggestions 😄

VocableUITests/Tests/SettingsScreenTests.swift
CananArikan: We disabled this test for now, the code you pointed out is something we plan to refactor but we are waiting for #492 to be completed because of the Categories UI changes. When that ticket is completed we plan to update the flow in this test.

VocableUITests/Screens/SettingsScreen.swift
CananArikan: We were thinking that we would use it outside of this context when we fix some of the other tests. But right now it is just being used here. I think we will go ahead and make it private. We can always change it back when we need to 😉

VocableUITests/Tests/CustomCategoriesTests.swift
Clstroud: Just realized this ^ may not actually be the accessibility ID, so I might've accidentally lied there. It's definitely the localization key, and I'm thinking we may want to make them match up for elements like this.

VocableUITests/Tests/SettingsScreenTests.swift
CananArikan: We try to “handle" what happens if the cell is not present initially in the loop within `locateCategoryCell()`. We don't throw a failure explicitly but we do let the query fail which will report that the cell could not be found. Before that, however, we scroll the pages within the list of categories to find the cell we're looking for. 

Making a reference to the cell once in order to shorten our run time and the line of code is a good idea. We will implement that.

VocableUITests/Tests/CustomCategoriesTests.swift
CananArikan: Fixed!

VocableUITests/Tests/SettingsScreenTests.swift
CananArikan: Fixed.

VocableUITests/Screens/SettingsScreen.swift
CananArikan: We updated it!

Clstroud (approved):
🏆 💯 🥇 Nice updates!

PR #521, opened 2022-03-29
Update Localizable.strings

SashankPatel:
# Description of Work
Update the empty state string to what it should be to see if the Crowdin Sync issues are resolved.

## Notes to Test (Optional)
N/A

---


Clstroud (approved):
💯

PR #530, opened 2022-03-31
#461 & #484 | Existing Phrase & Category Refactor

### LINKED ISSUE #461: Create/Edit Custom Category -  Existing Category Alert
As a user, I want to be notified when I am adding a custom category that matches a pre-existing category. This includes any of my other custom categories.

Acceptance Criteria:

If a user adds a category that's already been added before, we should present an alert notifying them: "This category already exists. Continue to add the category?" with the option of "Continue" and "Cancel".

Design:

[3.6  Settings_CategoriesPhrases_AddCategory](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable-New?node-id=1%3A1344)

Linked Issues:
This ticket was derived from: #390
This ticket blocks development of: #397  -> the AC from this ticket are required to complete 397

### LINKED ISSUE #484: Custom Category: Adding a Phrase - Alert When Duplicating Existing Phrase
**Story:** As a user, I want to be notified when I am adding a custom phrase that matches a pre-existing phrase. This includes any of my other custom phrase.

**Acceptance Criteria:**
1. If a user adds a phrase that's already been added before, we should present an alert notifying them: "This phrase already exists. Continue to add the phrase?" with the option of "Continue" and "Cancel".

Design: 
Discussed in meeting but need to follow up with Steve

jsmorgan42:
closes #461 & #484

# Description of Work
Refactoring the `EditTextViewController` into a `TextEditorViewController` which contains a `TextEditorConfigurationProviding` delegate that other configuration providers implement to handle the various types of text editing (adding favorites, editing phrases, and editing categories). This PR additionally adds an alert that notifies the user if they are adding a category that has a pre-existing name as well as an alert that notifies the user if they are adding a phrase that already exists in the category its contained within.

## Notes to Test 
- This refactor touches all instances of editing text through a keyboard, so it would be good to test all text editing flows including:
     - In the home screen when adding favorites to the `My Sayings` category
     - In the home screen when adding a new phrase to an empty category
     - In settings when adding a new category in the categories list 
     - In settings when editing an existing category name in the category detail screen contained within the categories list
     - In settings when adding a new phrase in the edit phrases list contained within the category detail screen
     - In setting when editing an existing phrase in the edit phrases list contained within the category detail screen

---


Vocable/Common/TextEditorViewController.swift
```diff
@@ -12,40 +12,35 @@ import UIKit
...
+        var leftItemConfiguraton: TextEditorNavigationButton.Configuration?
+        var rightItemConfiguration: TextEditorNavigationButton.Configuration?
     }
 
+    let textView = OutputTextView(frame: .zero)
+
+    let leftButton = TextEditorNavigationButton()
+    let rightButton = TextEditorNavigationButton()
+
+    var delegate: TextEditorConfigurationProviding?
```
moyerr, line 33: So it looks like this is leading to a retain cycle with the `FreeResponseEditorConfigurationProvider` since that one is a class.

rhsalas3:
@jsmorgan42 

@CananArikan and I have tested this PR branch and, while the alert appears and matches designs, there are some issues we think should be addressed before this can be signed off on.

Issues we're seeing:
- [x] The free form keyboard, from the main screen, no longer uses the star icon for adding phrases to the 'My Sayings' category. This seems to be a regression b/c we don't see it changed in designs. It is the checkmark icon.

https://user-images.githubusercontent.com/34660497/161617724-e0901caf-482c-4b0c-83f5-aaf5bea68bb2.mp4



- [x] When adding a phrase or category, regardless of flow, if a user opts to autocomplete with suggested phrases you do not get the duplication alert. For example: add category "Hello" by typing in each letter. Then attempt to add a duplicate by using the "Hello" that is suggested when typing. Note the extra space at the end of the entry. When saving, no alert is given for a duplicate.
    - [x] Using suggested entry for Category name
    - [x] Using suggested entry for phrase
    - [x] Saving suggested phrase entry to 'My Sayings' from free form keyboard; "Hello" and "Hello " should be recognized as the same when adding to 'My Sayings' (thus the star would indicate it is already saved, etc).

https://user-images.githubusercontent.com/34660497/161617481-0e8e7b63-929a-4354-ad9c-648f3d43d9f0.mp4



- [x] When deleting a preset Category or preset Phrase so that there are none like it remaining, you still are alerted to duplicates when re-adding the same category/phrase.
 
https://user-images.githubusercontent.com/34660497/161617941-cf5b50c0-8c0f-4913-a8cc-e7e594570980.mp4



- [x] Duplicate alert does not appear when adding a new category that matches a preset category with two word titles, because the second word does not match casing. Example: "Basic Needs" is a preset, but you do not get a warning for a duplicate category name when you add "Basic needs" because the casing does not match (keyboard does not allow case changing)

https://user-images.githubusercontent.com/34660497/161618172-39349566-8f74-4647-9dd3-eabf4e9b8210.mp4

Vocable/Common/Views/TextEditorNavigationButton.swift
```diff
@@ -0,0 +1,79 @@
...
+
+        static func save(isEnabled: Bool = false, accessibilityIdentifier: String = "keyboard.saveButton", action: @escaping Action) -> Self {
+            Configuration(image: UIImage(systemName: "checkmark")!,
+                          isEnabled: isEnabled,
+                          accessibilityIdentifier: accessibilityIdentifier,
+                          action: action)
+        }
+
+        static func favorite(image: UIImage, isEnabled: Bool = false, accessibilityIdentifier: String = "keyboard.favoriteButton", action: @escaping Action) -> Self {
+            Configuration(image: UIImage(systemName: "checkmark")!,
```
moyerr: Looks like you're accepting an image as a parameter, but not actually using it.

Vocable/Common/TextEditorViewController.swift
jsmorgan42: Turns out this was related to keeping a strong reference of the `TextEditorViewController` instance that we send back to the delegate, meaning there were also retain cycles for the other struct delegate providers. This was fixed by making the view controller explicitly `weak` in the capture list of all the closures in all of the providers.

jsmorgan42:
@rhsalas3 @CananArikan All of these issues should be fixed now!

rhsalas3:
@jsmorgan42 

We re-tested and everything looks good! All of the issues have been fixed and the alert behaves as expected. 🎉 🎉 

CC: @SashankPatel

Vocable/Common/TextEditorViewController.swift
```diff
@@ -73,27 +68,53 @@ class EditTextViewController: VocableViewController, UICollectionViewDelegate {
...
+        setNeedsUpdateConfiguration()
+    }
+
+    private func handleTextChange() {
+        keyboardViewController.$attributedText
+            .dropFirst()
+            .map { [weak self] attributedText -> NSAttributedString? in
+                self?.textView.attributedText = attributedText
+                return attributedText
+            }
```
moyerr, line 91: Is this not something that can be done in the `sink`? Feels strange to shoehorn this side-effect into a `map` when there is no actual mapping happening.

Vocable/Common/Views/TextEditorNavigationButton.swift
```diff
@@ -0,0 +1,80 @@
...
+
+        static func save(isEnabled: Bool = false, accessibilityIdentifier: String = "keyboard.saveButton", action: @escaping Action) -> Self {
+            Configuration(image: UIImage(systemName: "checkmark")!,
+                          isEnabled: isEnabled,
+                          accessibilityIdentifier: accessibilityIdentifier,
+                          action: action)
+        }
+
+        static func favorite(image: UIImage, isEnabled: Bool = false, accessibilityIdentifier: String = "keyboard.favoriteButton", action: @escaping Action) -> Self {
+            Configuration(image: image,
```
moyerr: If we're using the same favorite icons throughout, do you think it'd make more sense to accept an `isFavorited` flag and instantiate the appropriate image in here as an implementation detail?

Vocable/Features/Root/FreeResponseTextEditorConfigurationProvider.swift
```diff
@@ -0,0 +1,105 @@
...
+    func textEditorViewControllerConfiguration(_ viewController: TextEditorViewController) -> TextEditorViewController.Configuration {
+        let currentUtterance = viewController.text?.trimmingCharacters(in: .whitespacesAndNewlines)
+
+        if needsUpdateFavoritesState {
+            updateFavoriteState(for: currentUtterance)
+        }
+
+        let leftConfiguration = TextEditorNavigationButton.Configuration.dismissal(for: viewController)
+
+        let rightConfiguration = TextEditorNavigationButton.Configuration.favorite(image: favoriteImage, isEnabled: canFavoritePhrase) { [weak self, weak viewController] in
```
moyerr: Hell yeah 💯

Vocable/Features/Settings/EditPhrases/EditPhrasesViewController.swift
```diff
@@ -123,24 +123,10 @@ final class EditPhrasesViewController: PagingCarouselViewController, NSFetchedRe
...
-
-                ToastWindow.shared.presentEphemeralToast(withTitle: alertMessage)
-            } catch {
-                assertionFailure("Failed to save user generated phrase: \(error)")
-            }
-        }
+        let viewController = TextEditorViewController()
+        let context = NSPersistentContainer.shared.newBackgroundContext()
+        viewController.delegate = PhraseEditorConfigurationProvider(categoryIdentifier: category.objectID,
+                                                           context: context)
```
moyerr, line 129: Big fan of all this complexity being moved out of these view controllers! 🎖️

moyerr (approved):
Looks good to me! Dropping an approval with some minor feedback, so feel free to merge when it's ready ✅

Vocable/Common/Views/TextEditorNavigationButton.swift
jsmorgan42: Great idea!

Vocable/Common/TextEditorViewController.swift
jsmorgan42: The reason we did this is so we could pull out the string of the attributed text before we call `removeDuplicates()` since attributed strings can have the same text but different attributes and they would not be removed as duplicates even though they should be.

PR #535, opened 2022-04-04
[CI] Attempt only exporting to Crowdin from develop branch

Clstroud:
Experimental pipeline update to only export translations from the develop branch. Goal is to reduce noise and build failures for Crowdin PRs. More work may be needed.


SashankPatel (approved):
Ship it.

PR #546, opened 2022-04-06
#544 | Disable Heading Tracking Alert Copy Update

### LINKED ISSUE #544: Update Disable HeadTracking Alert Message
Currently when the user goes to settings > selection mode > hovers on head tracking there is an alert that says "Turn off headtracking?". We want to be more clear so that the user wont mistakenly turn it off and unable to use the app. 

**Acceptance Criteria**:

- Text is updated to "Turning off head tracking will make the app touch mode only. Would you like to continue?"
- Alert styling is consistent with other alerts on the app (exit without saving, removing categories, deleting phrases etc. )

**Design**:

- [Headtracking Alert Message](https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=460%3A1568)

jsmorgan42:
closes #544

# Description of Work
Updating copy for alert when disabling head tracking.

---

### DIFF
--- a/CrowdinExport/en.xliff
+++ b/CrowdinExport/en.xliff
@@ -158,13 +158,13 @@
         <note>Cancel alert action title</note>
       </trans-unit>
       <trans-unit id="gaze_settings.alert.disable_head_tracking_confirmation.button.confirm.title" xml:space="preserve">
-        <source>Confirm</source>
-        <target>Confirm</target>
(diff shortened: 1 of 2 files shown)


moyerr (approved):
💪

Clstroud (approved):
🚢

PR #550, opened 2022-04-07
#490 | Fixing Audio in Silent Mode 🔈 

### LINKED ISSUE #490: No Audio When Device is in 'Silent Mode'
***Description***
It is expected that phrases are read aloud when selected regardless of whether or not your device is in 'Silent Mode' (this is a device setting, not an app setting). On build 1.4.0-1760 running on an iPhone 8 Plus (iOS 14.8.1), while silent mode is on, none of the selected phrases are read aloud.

***To Reproduce***
1. Make sure that the 'Silent Mode' toggle switch is set to the OFF position. This is a physical switch on the left hand side of the device, just above the volume buttons.
2. Open the Vocable app
3. Navigate to any preset or custom phrase 
4. tap on the phrase > note that the phrase is read aloud (as expected)
5. Turn ON 'Silent Mode' using the same toggle switch in step 1 (the switch will have a red indicator to denote it is in the ON position)
6. Repeat steps 2 through 4 and note that the phrase _IS NOT_ read aloud.

***Expected behavior***
Regardless of the state of the 'Silent Mode' toggle switch on a device, the Vocable app will read aloud any selected phrases.

***Actual behavior***
The Vocable app will only read aloud selected phrases while 'Silent Mode' is OFF.

***Device Information***
 - Device: iPhone 8 Plus
 - OS: 14.8.1
 - Model #: MQ8D2LL/A

***Screenshot/Recording***

https://user-images.githubusercontent.com/34660497/159509607-75bc3e8e-39d1-45aa-a454-d5cf9d0cf7d7.MP4

jsmorgan42:
closes #490

# Description of Work
The audio was not playing in silent mode when tapping on phrases or reading out text. To fix this, the `AudioEngineCategory` is initialized as a property in `SpeechRecognitionController` so that the audio session category is initially set to `playback` to allow audio in silent mode. Also fixed issue where audio session wasn't switching back to `playback` after turning off listening mode.

---

### DIFF
--- a/Vocable/AppDelegate.swift
+++ b/Vocable/AppDelegate.swift
@@ -63,13 +63,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Warm up the speech engine to prevent lag on first invocation
         AVSpeechSynthesizer.shared.speak("", language: "en")
 
-        // Allows audio to be played in silent mode
-        do {
-            try AVAudioSession.sharedInstance().setCategory(.playback, mode: .spokenAudio, options: .duckOthers)
-        } catch let error {
-            print("Failed to set audio category: \(error.localizedDescription)")
-        }
-
         application.isIdleTimerDisabled = true
 
         NotificationCenter.default.addObserver(self,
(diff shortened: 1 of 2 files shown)


rhsalas3:
Tested and verified to be working as expected. Regardless of the state of the 'silent mode' switch on an iPhone device, the app will speak aloud selections. 

I did find an issue whereby the volume level of the output is very low when listening mode is engaged. It corrects itself when you turn off listening mode. But this behavior is reproducible on develop. So I will write up a separate bug for it.

moyerr (approved):

PR #557, opened 2022-04-11
#477 | Fixing category selection after deletion

### LINKED ISSUE #477: iOS - Unexpected view on Home Screen after deleting category.
**Describe the bug**
After a Category is deleted and I navigate to back to the main screen I expect to see the category immediately after the one that was deleted, but instead I see the one after that.

**To Reproduce**
The current order of categories is:
1.) General 
2.) Basic Needs
3.) Personal Care

1.  Launch the app - reinstall if needed.
2. Navigate to the General Category if not already
3. Navigate to Settings > Categories and Phrases.
4. Delete the General Category
5. Navigate out of settings and back to the Main/Home screen.

**Expected behavior**
As noted above: With the General category deleted I would expect Basic Needs category to be displayed on the Main screen.

**Actual behavior**
The Personal Care category was displayed.

**Device Information**
Run on an iPhone 8 simulator.  Should not be device dependent.

jsmorgan42:
closes #477

# Description of Work
Fixing a bug where after deleting the first category and navigating back to the home screen, the second category was selected inside of the first.

---

### DIFF
--- a/Vocable/Features/Root/CategoriesCarouselViewController.swift
+++ b/Vocable/Features/Root/CategoriesCarouselViewController.swift
@@ -194,6 +194,11 @@ import Combine
             }
 
             let newItemIndexPath: IndexPath? = {
+                if let category = Category.fetchObject(in: NSPersistentContainer.shared.viewContext, matching: previous.objectID),
+                   category.isUserRemoved {
+                    return previous.indexPath
+                }
+
                 if let indexPath = self.dataSourceProxy.indexPath(for: previous.objectID) {
                     return indexPath
                 }
--- a/Vocable/Features/Root/CategoriesCarouselViewController.swift
+++ b/Vocable/Features/Root/CategoriesCarouselViewController.swift
@@ -194,11 +194,6 @@ import Combine
             }
 
             let newItemIndexPath: IndexPath? = {
-                if let category = Category.fetchObject(in: NSPersistentContainer.shared.viewContext, matching: previous.objectID),
-                   category.isUserRemoved {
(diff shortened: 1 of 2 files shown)


moyerr (approved):

Vocable/Features/Root/CategoriesCarouselViewController.swift
```diff
@@ -194,6 +194,11 @@ import Combine
             }
 
             let newItemIndexPath: IndexPath? = {
+                if let category = Category.fetchObject(in: NSPersistentContainer.shared.viewContext, matching: previous.objectID),
+                   category.isUserRemoved {
+                    return previous.indexPath
+                }
```
Clstroud: I've been staring at this for a hot minute and I'm wondering if I'm misunderstanding something. `previous.indexPath` is no longer relevant to the new snapshot and could even be (I suppose?) out of bounds for said new snapshot, right?

What happens if you return `nil` instead?

Clstroud (changes requested):
Just one question I'd like to make sure is a non-issue before I unblock this thing. Otherwise, nice and simple fix 💯

rowildcard:
Test verified on:
simulator iPhone 8 Plus, iOS 15.4
iPhone 12 Pro Max, iOS 15.1

Clstroud (approved):
💯 Much safer. Good stuff 🥇

CananArikan:
Test verified on:
simulator iPhone 8, iOS 15.4
iPad Pro (11-inch) (3rd generation) 15.3.1

PR #558, opened 2022-04-11
#553 | Fixing Volume Issue During Listening Mode 🎛️ 

### LINKED ISSUE #553: Volume Level Issues During Listening Mode
## Describe the bug
On an iPhone device, while engaged in listening mode the output volume is too low to be audible regardless of using the listening mode suggestions or any of the preset phrases. In this state you can turn up the device volume until it is audible or turning off listening mode reverts the output level back to an audible level. However, considering our target user base, it may be troublesome to require a user to turn the volume up when they are physically unable.

**Steps To Reproduce:**
1. Open the app and tap on a phrase, note the volume level
1. Navigate to the 'Listen' category
2. Engage in listening mode; grant access to speech recognition and to the microphone
3. Navigate to any other category, and tap on a phrase. 
4. Note the output volume
5. Swipe open the device's 'Command Center' to adjust the volume
6. Return to Vocable and tap a phrase, note the volume is normal but quickly returns to being inaudible.

**Expected behavior**
The audio output level of the app while in listening mode is audible/adheres to the device's volume level at the time listening mode is engaged.

**Actual behavior**
The audio output level in listening mode defaults to being too low to be heard and must be adjust by the device's volume controls.

## Screenshots/Recordings

https://user-images.githubusercontent.com/34660497/162472376-27405099-7a02-4515-bfee-b2a7c967b1ed.MOV

## Device Information

 - Device: iPhone 8 Plus
 - OS: iOS 14.8.1
 - CPU: [A11 Bionic](https://www.apple.com/am/iphone-8/specs/)

jsmorgan42:
closes #553 

# Description of Work
This PR fixes a bug where the output volume was too low for listening mode which was fixed by setting the audio mode to `spokenAudio` instead of `voiceChat`. This mode seems to lower the output volume by default. Apple states that the `voiceChat` mode should only be used for "two-way voice communication, such as using Voice over Internet Protocol (VoIP)" which shouldn't be applicable in our case since it's used for text to speech.

---

### DIFF
--- a/Vocable/Features/Voice/AudioEngineController.swift
+++ b/Vocable/Features/Voice/AudioEngineController.swift
@@ -181,7 +181,7 @@ class AudioEngineController: NSObject, AVAudioPlayerDelegate {
                 if self.registeredSpeechControllers.isEmpty {
                     if self.audioSession.category != .playback {
                         print("AUDIO SESSION CATEGORY: playback")
-                        try self.audioSession.setCategory(.playback, mode: .spokenAudio, options: .interruptSpokenAudioAndMixWithOthers)
+                        try self.audioSession.setCategory(.playback, mode: .spokenAudio, options: .duckOthers)
                         sessionNeedsActivation = true
                     }
                     self.audioEngineShouldRun = !self.registeredSpeechControllers.isEmpty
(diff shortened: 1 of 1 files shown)


moyerr (approved):
🗣

Clstroud (approved):
Messed around with it and I agree that this seems to be working okay. The category/mode options can lead to wildly different behaviors with how recording and playback are handled, how things get mixed, etc. 

In earlier iterations I had issues with e.g. the device hearing itself, but I think that type of thing has been mitigated by dynamically managing these configurations.

rhsalas3:
Tested and verified to be working as expected. I also did some lite regression on iPad. Audio volume is good throughout the app when listening mode is on for the iPhone device I have, including while the 'silent mode' of the device is on.

👍  🥳 🎉

PR #560, opened 2022-04-12
Fixed Settings Screen Tests

CananArikan:
testReorder is fixed and enabled.
The parent issue is https://github.com/willowtreeapps/vocable-ios/issues/405

# Description of Work
Fixed broken testReorder test. This is part of the effort to unlink the chained tests.


Clstroud (approved):
I have some vague ideas about how we could make this more expressive in the future, but I don't have any notes about your implementation as-is. Good job @CananArikan 😄 💯

PR #564, opened 2022-04-13
Pagination UI Tests: Fix and Refactor

### LINKED ISSUE #470: Handle/Fix Pagination Tests
Pagination tests are failing. These tests need to be refactored to run independent of other tests (part of the "unlink tests" work) and pass.

This issue is a subtask of #405 

**Acceptance Criteria**:

All pagination tests pass.

rhsalas3:
closes #470 

## Description
Originally there were pagination tests that were very long and cumbersome. So I moved them from their original suite Class and added the tests to their own, splitting the larger tests into smaller tests to reduce overhead and potential points of failure. Also, since I was under the hood anyway, I refactored these tests to pass on iPad devices as well. There are also helper functions that did not work because of the new category details screen, so I refactored them.

---

## Notes to Test
The `CategoryPhrasesPaginationTests` and `MainScreenPaginationTests` are the new tests. They, as well as all of the others, should pass on an iPhone and iPad device.


Vocable/Features/Settings/EditCategories/EditCategoriesViewController.swift
```diff
@@ -109,6 +111,7 @@ final class EditCategoriesViewController: PagingCarouselViewController, NSFetche
         }
 
         cell.contentConfiguration = config
+        cell.accessibilityIdentifier = category.identifier
```
rhsalas3, line 114: My intent here is to be able to craft a unique query indifferent of duplicate categories, or different language settings. By adding the `category`'s identifier to the cell and an identifier on the button I can do something like `app.cells[<customCategory_ID>].buttons["edit_category_button"]`.

VocableUITests/Screens/CustomCategoriesScreen.swift
```diff
@@ -26,17 +26,45 @@ class CustomCategoriesScreen: BaseScreen {
         keyboardScreen.checkmarkAddButton.tap()
     }
     
-    func addCustomPhrases(numberOfPhrases: Int) {
+    func createAndLocateCustomCategory(_ categoryName: String) -> CategoryIdentifier {
+        createCustomCategory(categoryName: categoryName)
+        let customCategoryIdentifier = settingsScreen.locateCategoryCell(categoryName).element.identifier
+        
+        return CategoryIdentifier(customCategoryIdentifier)
+    }
```
rhsalas3: This allows me to use the returned `CategoryIdentifier` to locate the custom category on the Main Screen.

VocableUITests/Screens/SettingsScreen.swift
```diff
@@ -120,19 +117,8 @@ class SettingsScreen: BaseScreen {
...
-            leaveCategoriesButton.tap()
-            exitSettingsButton.tap()
-        case "settings":
-            exitSettingsButton.tap()
-        default:
-            break
-        }
+    func returnToMainScreen() {
+        navBarDismissButton.tap()
     }
```
rhsalas3: The thought here was to replace this original `navigateToMainScreenFromSettings(String)` function with a function in each screen class.

VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
```diff
@@ -0,0 +1,76 @@
...
+
+import XCTest
+
+class CategoryPhrasesPaginationTests: CustomPhraseBaseTest {
+    
+    func testCanNavigatePages() {
+        // Add 11 phrases; this results in 3 total pages for an iPhone and 2 total pages for an iPad
+        for phrase in listOfPhrases.startIndex..<11 {
+            customCategoriesScreen.addPhrase(listOfPhrases[phrase])
+        }
```
rhsalas3: In order to get the pagination tests to pass on both device types, I had to find the "magic number" of phrases; iPad displays 10 phrases, iPhone - 4 phrases

VocableUITests/Tests/CustomPhraseBaseTest.swift
```diff
@@ -11,14 +11,19 @@ import XCTest
 class CustomPhraseBaseTest: BaseTest {
     
     private(set) var customCategoryName: String = "Test"
+    private(set) var customCategoryIdentifier: CategoryIdentifier?
+    
+    // To avoid potential duplication from random strings, we'll have our own phrase bank to pull from.
+    private(set) var listOfPhrases: [String] = ["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P"]
```
rhsalas3, line 17: Every 3rd run or so, using random phrases (even at 2 chars in length) would duplicate because of the sheer number needed. So this way I can guarantee what phrases are used in the tests.

VocableUITests/Screens/CustomCategoriesScreen.swift
```diff
@@ -26,17 +26,45 @@ class CustomCategoriesScreen: BaseScreen {
...
-            // Do this until accessibility identifiers are add and trash icon can be accessed.
-            // Not needed until when change the screen to stay.
-            //keyboardScreen.dismissKeyboardButton.tap()
         }
     }
     
+    func returnToMainScreenFromCategoriesList() {
+        navBarBackButton.tap()
+        _ = navBarDismissButton.waitForExistence(timeout: 0.25)
+        navBarDismissButton.tap()
```
jsmorgan42: I wonder if we could add a helper function within a `XCUIElement` extension called something like `waitAndTap()` that'll waitForExistence and then tap the element. Could also add a timeout parameter to it with a default of `0.25`.

VocableUITests/Screens/CustomCategoriesScreen.swift
```diff
@@ -26,17 +26,45 @@ class CustomCategoriesScreen: BaseScreen {
...
+        _ = mainScreen.settingsButton.waitForExistence(timeout: 0.25)
+    }
+    
+    func returnToMainScreenFromCategoryDetails() {
+        navBarBackButton.tap()
+        _ = navBarBackButton.waitForExistence(timeout: 0.25)
```
jsmorgan42: Wouldn't we want the `waitForExistence()` before we tap the button?

VocableUITests/Screens/BaseScreen.swift
```diff
@@ -13,9 +13,14 @@ import Foundation
...
 
 class BaseScreen {
+    
+    let navBarBackButton = XCUIApplication().buttons["navigationBar.backButton"]
     let paginationLabel = XCUIApplication().staticTexts["bottomPagination.pageNumber"]
+    let paginationLeftButton = XCUIApplication().buttons["bottomPagination.left_chevron"]
+    let paginationRightButton = XCUIApplication().buttons["bottomPagination.right_chevron"]
     let alertMessageLabel = XCUIApplication().staticTexts["alert_message"]
     let emptyStateAddPhraseButton = XCUIApplication().buttons["empty_state_addPhrase_button"]
+    let navBarDismissButton = XCUIApplication().buttons.containing(NSPredicate(format: "identifier CONTAINS 'dismissButton'")).element
```
jsmorgan42: Love moving all these generic UI elements into the BaseScreen 💯

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    func testAddingPhrasesAdjustsPagination() {
+        // Use array slices to split the phrase bank into two sets of 8 phrases
+        let listMidpoint = listOfPhrases.count / 2
+        let firstSetOfPhrases = listOfPhrases[..<listMidpoint]
+        let secondSetOfPhrases = listOfPhrases[listMidpoint...]
+        
+        // Add the first 8 phrases
+        for phrase in firstSetOfPhrases.startIndex..<firstSetOfPhrases.endIndex {
+            customCategoriesScreen.addPhrase(firstSetOfPhrases[phrase])
+        }
```
jsmorgan42: Not necessary to change but you can make this a bit more swift-y by leveraging the `forEach` array function so it would look something like this instead:

```
        firstSetOfPhrases.forEach { phrase in
            customCategoriesScreen.addPhrase(phrase)
        }
```

Looks like there's a few other places where this could be used as well.

rowildcard:
All of the tests passed locally so if you address a couple of Jesse's comments I think you'll be good to go.

VocableUITests/Screens/CustomCategoriesScreen.swift
rhsalas3: That would be useful! Thank you for the suggestion, I'll look into it.

VocableUITests/Screens/CustomCategoriesScreen.swift
rhsalas3: I am waiting for the `navBarBackButton` on the subsequent screen, before calling `returnToMainScreenFromCategoriesList()`.

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: This is neat, I will implement this.

rhsalas3: Done!

VocableUITests/Screens/CustomCategoriesScreen.swift
```diff
@@ -26,17 +26,45 @@ class CustomCategoriesScreen: BaseScreen {
         keyboardScreen.checkmarkAddButton.tap()
     }
     
-    func addCustomPhrases(numberOfPhrases: Int) {
+    func createAndLocateCustomCategory(_ categoryName: String) -> CategoryIdentifier {
+        createCustomCategory(categoryName: categoryName)
+        let customCategoryIdentifier = settingsScreen.locateCategoryCell(categoryName).element.identifier
+
```
Clstroud:
```suggestion
```

VocableUITests/Screens/CustomCategoriesScreen.swift
Clstroud: ^ How are you differentiating the buttons between screens? In other words, what prevents the test runner from erroneously grabbing the navBarBackButton on the previous screen and moving on during a slow transition?

VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
Clstroud: Can/should we make it less magical by pulling that logic into a place where it can be referenced from the test runner? Thinking `UIDevice.current.userInterfaceIdiom` should do the trick, but this may also be orientation specific.

VocableUITests/Tests/MainScreenPaginationTests.swift
Clstroud: I think you _might_ be able to use [`Array.stride`](https://developer.apple.com/documentation/swift/1641347-stride) to paginate your array and then loop through the result? As best I can tell everything is duplicated for both pages except for the page number, which could be handled using the `.indexed` or `.enumerated()` function.

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        
+        // Return to the Main Screen and navigate to the test category
+        customCategoriesScreen.returnToMainScreenFromEditPhrases()
+        mainScreen.locateAndSelectDestinationCategory(customCategoryIdentifier!)
+        
+        // Verify that the category's pagination is "Page 1 of 2"; page next buttons are enabled
+        XCTAssertEqual(mainScreen.currentPageNumber, 1)
+        XCTAssertEqual(mainScreen.totalPageCount, 2)
+        XCTAssertTrue(mainScreen.paginationLeftButton.isEnabled)
+        XCTAssertTrue(mainScreen.paginationRightButton.isEnabled)
```
Clstroud: I'm seeing a _lot_ of these pagination of checks, and I'm thinking a [custom assert](https://www.swiftbysundell.com/articles/test-assertions-in-swift/) may cut down on a lot of repetitive code and make your call sites a lot clearer

Clstroud: For example:
```
    func WTAssertPaginationEquals(
        _ currentPage: Int,
        of numberOfPages: Int,
        leftArrowEnabled: Bool = true,
        rightArrowEnabled: Bool = true,
        file: StaticString = #file,
        line: UInt = #line
    ) {
        XCTAssertEqual(mainScreen.currentPageNumber, 1, file: file, line: line)
        XCTAssertEqual(mainScreen.totalPageCount, 2, file: file, line: line)
        XCTAssertTrue(mainScreen.paginationLeftButton.isEnabled, file: file, line: line)
        XCTAssertTrue(mainScreen.paginationRightButton.isEnabled, file: file, line: line)
    }
```

Clstroud: e.g. `WTAssertPaginationEquals(1, of: 3)` etc

Clstroud (approved):
Good stuff! 💯 Left a few thoughts -- definitely look into that assertion if you haven't already 😄

VocableUITests/Screens/CustomCategoriesScreen.swift
rhsalas3: Done!

VocableUITests/Screens/CustomCategoriesScreen.swift
rhsalas3: I refactored these methods to use your suggested XCUIElement extension, `waitAndTap()`. So maybe this'll be more intuitive with the new implementation.

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: I like this, I'll take some time to implement.

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: So instead of creating two separate array slices,  I can use `Array.stride` to get similar functionality? 🤔 

I will look into it.

VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
rhsalas3: It is orientation specific, but then there could just be one for each orientation...maybe?

~~I'm not sure how it would be helpful in this particular test's case, since I got the number 11 by walking through my test steps on the device and ensuring that number worked on both. I would think we would want the `userInterfaceIdiom` value to always be the number of entries on the page. So I would still have to manipulate that number to get to 11~~...Oh, ok, strike that. That still achieves your intent, "make it less magical". I get it better now having thought through it more while typing out this response lol

I like understandable/fathomable in code better than 🧙  magical 🧙  so I'll take a shot at implementing this lol

rhsalas3: 🤔  I tried to implement something around your suggestion that would make sense for the needs in my tests, around the `UIDevice.current.userInterfaceIdiom` value, like this:
```
var maxPhrasesInPortrait: Int {
        if (UIDevice.current.userInterfaceIdiom == .pad) {
            return 10
        } else {
            return 4
        }
    }
```
...but the numbers I need in the tests are the same regardless of interface. So that led to additional code to check the usertInterfaceIdiom again and add values to the returned value from the above property, so that it was the same number regardless of device type. For example: `if (.pad) ?  maxPhrasesInPortrait + 1 : maxPhrasesInPortrait + 4`; saying "hey, if you're iPad give me 11...and if your iPhone give me 11" didn't feel like it was helping. 

I'll keep `UIDevice.current.userInterfaceIdiom` in mind for when we need the exact number of phrases on the screen. Thank you for the suggestion.

VocableUITests/Screens/CustomCategoriesScreen.swift
rhsalas3: Fixed

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: After implementing what I thought you were suggesting...
```
for phrase in stride(from: listOfPhrases.startIndex, to: listMidpoint, by: 1) {
    customCategoriesScreen.addPhrase(listOfPhrases[phrase])
}
```
...I felt like it wasn't what you really meant. I could not find anything online to help me. I think I'll keep my previous changes, and keep the stride function in my back pocket as a tool for future use.

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: @Clstroud 

I implemented a custom assertion based on your suggestion. I pushed it up with the latest commit(s) to this PR branch. Please take a look when you get a chance. Thanks!

VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
Clstroud: Might be a good candidate for an enhancement ticket because I think we can still make that better, but it needn't block this PR

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -0,0 +1,136 @@
...
+//
+
+import XCTest
+
+class MainScreenPaginationTests: CustomPhraseBaseTest {
+    
+    // In order to ensure this test passes on both an iPhone and iPad device, we will add 16 phrase (2 pages on an iPad)
+    // then remove 8 of them so that the pages reduce from 2 to 1 page.
+    func testDeletingPhrasesAdjustsPagination() {
+        listOfPhrases.forEach{ phrase in
```
jsmorgan42:
```suggestion
        listOfPhrases.forEach { phrase in
```

This would technically be a swiftlint warning without the spacing after the `forEach` declaration

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -0,0 +1,136 @@
...
+    }
+    
+    // In order to ensure there are 2 pages on iPad devices, we'll need 16 total phrases.
+    func testAddingPhrasesAdjustsPagination() {
+        // Use array slices to split the phrase bank into two sets of 8 phrases
+        let listMidpoint = listOfPhrases.count / 2
+        let firstSetOfPhrases = listOfPhrases[..<listMidpoint]
+        let secondSetOfPhrases = listOfPhrases[listMidpoint...]
+        
+        firstSetOfPhrases.forEach{ phrase in
```
jsmorgan42:
```suggestion
        firstSetOfPhrases.forEach { phrase in
```

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -0,0 +1,136 @@
...
+        // Navigate to main screen to verify page numbers; expected to be "Page 1 of 1"
+        customCategoriesScreen.returnToMainScreenFromEditPhrases()
+        mainScreen.locateAndSelectDestinationCategory(customCategoryIdentifier!)
+        VocableAssert().paginationEquals(1, of: 1, leftArrowEnabled: false, rightArrowEnabled: false)
+        
+        // Add 8 more phrases to verify that the pagination adjusts as expected; "Page 1 of 2"
+        settingsScreen.navigateToSettingsCategoryScreen()
+        settingsScreen.openCategorySettings(category: customCategoryName)
+        customCategoriesScreen.editCategoryPhrasesButton.tap()
+        secondSetOfPhrases.forEach{ phrase in
```
jsmorgan42:
```suggestion
        secondSetOfPhrases.forEach { phrase in
```

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -0,0 +1,136 @@
...
+        }
+        
+        customCategoriesScreen.returnToMainScreenFromEditPhrases()
+        mainScreen.locateAndSelectDestinationCategory(customCategoryIdentifier!)
+        VocableAssert().paginationEquals(1, of: 2)
+    }
+    
+    func testCanScrollPagesWithPaginationArrows() {
+        // Add enough phrases to push the total number of pages to at leaset 2 for iPad and iPhone (16).
+        listOfPhrases.forEach{ phrase in
```
jsmorgan42:
```suggestion
        listOfPhrases.forEach { phrase in
```

VocableUITests/XCUIElement.swift
```diff
@@ -0,0 +1,27 @@
...
+
+extension XCUIElement {
+    
+    /// Waits the amount of time you specify for the element’s exists property to become true and then taps on it.
+    ///
+    /// Returns false if the timeout expires without the element coming into existence. In this case, the `tap` action
+    /// will not occur.
+    func waitForThenTap(timeout: TimeInterval = 0.25) -> Bool {
+        let elementDidAppear = self.waitForExistence(timeout: timeout)
+        if (elementDidAppear) {
```
jsmorgan42:
```suggestion
        if elementDidAppear {
```

In swift, there's no need for parentheses around conditionals in an `if` statement 🙂

jsmorgan42 (approved):
🚀 🔥

VocableUITests/Tests/MainScreenPaginationTests.swift
Clstroud: Well, I think I conflated some stuff but here's another stab at cleaning up the part I wasn't very excited about originally:
```
        let itemsPerPage = listOfPhrases.count / 2
        let paginatedPhrases = Array(listOfPhrases.chunks(ofCount: itemsPerPage))
        let pageOnePhrases = paginatedPhrases[0]
        let pageTwoPhrases = paginatedPhrases[1]
```
To do this you'd have to import Swift Algorithms at the top with `import Algorithms` (we already use it in the main app) and make sure the test target links against it in the build settings.

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -0,0 +1,136 @@
...
+            customCategoriesScreen.addPhrase(phrase)
+        }
+        
+        // Return to the Main Screen and navigate to the test category
+        customCategoriesScreen.returnToMainScreenFromEditPhrases()
+        mainScreen.locateAndSelectDestinationCategory(customCategoryIdentifier!)
+        
+        // Verify that the category's pagination is "Page 1 of 2"; page next buttons are enabled
+        VocableAssert().paginationEquals(1, of: 2)
```
Clstroud: I don't think you need to instantiate a `VocableAssert` every time you use one of these asserts 😄 I'll comment on the class itself to elaborate

VocableUITests/VocableAssert.swift
```diff
@@ -0,0 +1,33 @@
...
+//  Custom assertions used within the VocableUITests module.
+//
+//  Created by Rudy Salas on 4/18/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import Foundation
+import XCTest
+
+class VocableAssert: XCTest {
```
Clstroud: I think for the sake of brevity you may prefer to have these asserts be global functions with a sane prefix like `VOCAssert` or `VTAssert` for "Vocable Assert" or "Vocable Test Assert". To do that you would just declare them without any class or anything wrapping them.

If not, I would encourage you to use a case-less enum because there isn't really a need to have this exist as a dedicated type beyond _at most_ name-spacing (debatable). In this current version that would look like: `enum VocableAssert` and any assertions inside would be static members.

VocableUITests/VocableAssert.swift
```diff
@@ -0,0 +1,33 @@
...
+    
+    /// Use this function to confirm the expected state of the pagination label and its controls.
+    ///
+    /// By default this function expects the state of the left pagination arrow and the right pagination arrow to be isEnabled == true.
+    func paginationEquals(_ expectedCurrentPageNumber: Int,
+                          of expectedTotalPageCount: Int,
+                          leftArrowEnabled: Bool = true,
+                          rightArrowEnabled: Bool = true,
```
Clstroud: Seeing this in practice felt a little unnecessarily verbose, so here's one change you might prefer:

By adding an OptionSet such as this:

```
struct PaginationArrows: OptionSet {

    let rawValue: Int8

    static let left = PaginationArrows(rawValue: 1)
    static let right = PaginationArrows(rawValue: 1 << 1)

    static let both: PaginationArrows = [.left, .right]
    static let none: PaginationArrows = []
}
```
You could make your function thus:
```
func paginationEquals(_ 
    expectedCurrentPageNumber: Int,
    of expectedTotalPageCount: Int,
    enabledArrows: PaginationArrows = .both,
    file: StaticString = #file,
    line: UInt = #line
){
         XCTAssertEqual(BaseScreen().currentPageNumber, expectedCurrentPageNumber, file: file, line: line)
         XCTAssertEqual(BaseScreen().totalPageCount, expectedTotalPageCount, file: file, line: line)
         XCTAssertEqual(BaseScreen().paginationLeftButton.isEnabled, enabledArrows.contains(.left), file: file, line: line)
         XCTAssertEqual(BaseScreen().paginationRightButton.isEnabled, enabledArrows.contains(.right), file: file, line: line)
     }
```

And then your call site goes from:
```
VocableAssert().paginationEquals(1, of: 1, leftArrowEnabled: false, rightArrowEnabled: false)
```
to:
```
VocableAssert().paginationEquals(1, of: 1, enabledArrows: .none)
VocableAssert().paginationEquals(1, of: 1, enabledArrows: .left)
VocableAssert().paginationEquals(1, of: 1, enabledArrows: .right)
VocableAssert().paginationEquals(1, of: 1, enabledArrows: .both)
VocableAssert().paginationEquals(1, of: 1, enabledArrows: [.left, .right])
... etc
```

VocableUITests/VocableAssert.swift
```diff
@@ -0,0 +1,33 @@
...
+                          file: StaticString = #file,
+                          line: UInt = #line)
+    {
+        
+        XCTAssertEqual(BaseScreen().currentPageNumber, expectedCurrentPageNumber, file: file, line: line)
```
Clstroud: Does BaseScreen need to be instantiated every single time? I feel like all such screens should have their members defined statically since an instance of them doesn't really give you anything right?

VocableUITests/XCUIElement.swift
```diff
@@ -0,0 +1,27 @@
...
+import Foundation
+import XCTest
+
+extension XCUIElement {
+    
+    /// Waits the amount of time you specify for the element’s exists property to become true and then taps on it.
+    ///
+    /// Returns false if the timeout expires without the element coming into existence. In this case, the `tap` action
+    /// will not occur.
+    func waitForThenTap(timeout: TimeInterval = 0.25) -> Bool {
```
Clstroud: Naming niggle here: you're probably aiming for brevity over the system's function name, but I don't love that this name sorta leaves you hanging about what it's actually waiting for.

I suggest perhaps something along the lines of these:
```
@discardableResult
    func tap(afterWaitingForExistenceWithTimeout timeout: TimeInterval) -> Bool {
        return true
    }

    @discardableResult
    func tap(waitForExistence timeout: TimeInterval) -> Bool {
        return true
    }
```
Either way I think it overloads the existing tap function unambiguously if the default parameter is omitted (since that's implicitly a wait of 0 anyway)

VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
rhsalas3: Ok. I wrote up an enhancement ticket, so we can revisit this idea in the future.

#579

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: 😅

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: Fixed [all of them]

VocableUITests/VocableAssert.swift
rhsalas3: I like the idea of them being global functions. I'm almost certain now that that's how the XCTest XCTAssert functions are anyway, and what I wanted to do (but didn't know how) lol. 

Is it ok to still have them in a separate swift file? Otherwise, the next best place would be the `BaseTest` file. Thoughts?

VocableUITests/XCUIElement.swift
rhsalas3: Done

VocableUITests/VocableAssert.swift
rhsalas3: If I'm understanding you correctly, you're saying I should do this instead?
```
let baseScreen = BaseScreen()

func paginationEquals(_
    //...

    XCTAssertEqual(baseScreen.currentPageNumber, expectedCurrentPageNumber, file: file, line: line)
    XCTAssertEqual(baseScreen.totalPageCount, expectedTotalPageCount, file: file, line: line)
    // etc..
}
```

You do not need to instantiate BaseScreen each time. I just did that because I didn't think it all the way through. I just had the first thought, that "I need BaseScreen".

VocableUITests/VocableAssert.swift
rhsalas3: Updated!

VocableUITests/VocableAssert.swift
rhsalas3: I added the OptionSet and pushed up the other changes we made.

VocableUITests/VocableAssert.swift
rhsalas3: We discussed this in a Slack huddle and I implemented the changes so that the screen classes are defined statically.

VocableUITests/XCUIElement.swift
rhsalas3: Updated!

VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
```diff
@@ -0,0 +1,72 @@
...
+    func testCanNavigatePages() {
+        // Add 11 phrases; this results in 3 total pages for an iPhone and 2 total pages for an iPad
+        for phrase in listOfPhrases.startIndex..<11 {
+            CustomCategoriesScreen.addPhrase(listOfPhrases[phrase])
+        }
+        
+        // Verify that the user is on the first page and the next page buttons are enabled.
+        XCTAssertEqual(CustomCategoriesScreen.currentPageNumber, 1)
+        XCTAssertTrue(CustomCategoriesScreen.paginationLeftButton.isEnabled)
+        XCTAssertTrue(CustomCategoriesScreen.paginationRightButton.isEnabled)
```
Clstroud, line 22: I think this can use the new assert

VocableUITests/Tests/MainScreenTests.swift
```diff
@@ -64,111 +64,45 @@ class MainScreenTests: BaseTest {
...
+        SettingsScreen.navBarBackButton.tap()
+        SettingsScreen.navBarDismissButton.tap()
         
         // Confirm that the category is no longer accessible.
         for category in PresetCategories().list {
             // If we come across the category we expect to be hidden, fail the test. Otherwise the test will pass.
-            mainScreen.locateAndSelectDestinationCategory(category.categoryIdentifier)
-            if (mainScreen.selectedCategoryCell.identifier == hiddenCategoryIdentifier) {
+            MainScreen.locateAndSelectDestinationCategory(category.categoryIdentifier)
+            if (MainScreen.selectedCategoryCell.identifier == hiddenCategoryIdentifier) {
```
Clstroud:
```suggestion
            if MainScreen.selectedCategoryCell.identifier == hiddenCategoryIdentifier {
```

Clstroud (approved):
I think maybe one more pass to make sure you're using the pagination assert in all applicable places might be good, but this looks good! Nice work

VocableUITests/Tests/MainScreenTests.swift
rhsalas3: Fixed

PR #574, opened 2022-04-18
[488] Add head tracking support footer text to settings ui

### LINKED ISSUE #488: Head Tracking Does Not Work On iPhone 8 Plus - iOS 14.8.1
**Describe the bug**
It is expected that Apple devices that are able to run iOS 14.x will be able to utilize head tracking for the app. With an iPhone 8 Plus (model #MQ8D2LL/A) that has an A11 Bionic processor, the Vocable app does not allow for head tracking.

The version of the app under test is 1.4.0-1760 (from TestFlight).

Update: Could be the source of the issue stems from the device not asking for permission to use the camera. On a fresh install, it is expected that Vocable asks for camera access, but this is not happening on the iPhone 8 Plus under test.

**To Reproduce**
1. Install and launch the Vocable app.
2. Note that, at launch, there is no alert prompting for camera access.
3. From the Main Screen > tap the Settings button
4. tap on 'Selection Mode' 
5. tap on the 'Head Tracking' toggle switch

**Expected behavior**
The 'Head Tracking' toggle switch should be enabled, and tapping it will toggle on/off the head tracking feature.

**Actual behavior**
The 'Head Tracking' toggle switch is disabled.

**Device Information**
 - Device: iPhone 8 Plus
 - OS: 14.8.1
 - Model #: MQ8D2LL/A

Neural Engine Version History -> https://apple.fandom.com/wiki/Neural_Engine
iPhone Devices w/ Neural Engines -> https://apple.fandom.com/wiki/List_of_Apple_processors#A_series

Clstroud:
closes #488

# Description of Work
* Adds footer text to the head tracking settings screen if the device does not support head tracking:

Simulator Screen Shot - iPhone 8 - 2022-04-18 at 14 59 55

## Notes to Test (Optional)
* Devices that do support head tracking will not show this footer text

### DIFF
--- a/Vocable/Features/Settings/SelectionMode/SelectionModeViewController.swift
+++ b/Vocable/Features/Settings/SelectionMode/SelectionModeViewController.swift
@@ -176,12 +176,11 @@ final class SelectionModeViewController: VocableCollectionViewController {
         let systemVersion = UIDevice.current.systemVersion
         let sensorName = "TrueDepth"
         let neuralEngineName = "Apple Neural Engine"
-        let supportedPlatformsiOS14 = "iOS 14 and iPadOS 14"
 
         let format = NSLocalizedString("settings.selection_mode.head_tracking_unsupported_footer",
-                                       comment: "Footer text explaining that the user's device and/or operating system version is incompatible with head tracking and which devices and operating system versions support head tracking")
+                                       comment: "Footer text explaining that the user's device and/or operating system version is incompatible with head tracking and which devices do support head tracking")
 
-        let text = String(format: format, model, systemName, systemVersion, sensorName, neuralEngineName, supportedPlatformsiOS14)
+        let text = String(format: format, model, systemName, systemVersion, sensorName, neuralEngineName)
         return text
     }
 
--- a/Vocable/Features/Settings/SelectionMode/SelectionModeViewController.swift
+++ b/Vocable/Features/Settings/SelectionMode/SelectionModeViewController.swift
@@ -14,6 +14,10 @@ final class SelectionModeViewController: VocableCollectionViewController {
         case headTrackingToggle
     }
 
+    private enum SupplementaryKind: String {
+        case headTrackingUnsupportedFooter
+    }
+
     private lazy var dataSource: UICollectionViewDiffableDataSource<Int, SelectionModeItem> = .init(collectionView: collectionView) { (collectionView, indexPath, item) -> UICollectionViewCell in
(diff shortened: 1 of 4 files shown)


rhsalas3:
Tested and confirmed the footer text appears as expected, both on an iPhone and iPad device, when the head tracking feature is unsupported (iPhone 8 Plus - TD-706, iPhone 12 - Simulator, iPad 9th Gen - Simulator). 

I have also tested and confirmed that the footer text is **not** present when the device does support head tracking; toggling the feature on/off continues to works as before.

This fix looks good to me!

Vocable/Features/Settings/SelectionMode/SelectionModeViewController.swift
```diff
@@ -135,4 +167,22 @@ final class SelectionModeViewController: VocableCollectionViewController {
...
+        let systemName = UIDevice.current.systemName
+        let systemVersion = UIDevice.current.systemVersion
+        let sensorName = "TrueDepth"
+        let neuralEngineName = "Apple Neural Engine"
+        let supportedPlatformsiOS14 = "iOS 14 and iPadOS 14"
+
+        let format = NSLocalizedString("settings.selection_mode.head_tracking_unsupported_footer",
+                                       comment: "Footer text explaining that the user's device and/or operating system version is incompatible with head tracking and which devices and operating system versions support head tracking")
+
+        let text = String(format: format, model, systemName, systemVersion, sensorName, neuralEngineName, supportedPlatformsiOS14)
```
jsmorgan42: Not sure if it's relevant in this case but I've been told in the past that using `String.localizedStringWithFormat` is generally better practice for localization when it comes to formatted strings since it passes in the device locale by default

jsmorgan42 (approved):
Have a minor suggestion, otherwise 💯

Vocable/Features/Settings/SelectionMode/SelectionModeViewController.swift
Clstroud: For certain kinds of formatting I would agree, but we're specifically grabbing the localized version of the format string itself and _then_ building it like a normal formatted string. What you're suggesting would be great for things like fractional number formatting, where you're giving it a formatted value that you want to adapt to the user's locale: https://developer.apple.com/documentation/foundation/nsstring/1497301-localizedstringwithformat

The way this implementation is doing it allows translators to reorder the format specifiers for their locale as needed while also providing the translated strings.

jsmorgan42: Gotcha, that makes sense!

Issue #573, opened 2022-04-18
Hover Time UI tests


rowildcard:
Write unit tests for [test case TC-6](https://vocable.wtadev.com/case/6/): verify the Hover Time controls work. 

- User can decrease the Hover Time to the minimum allowed.
- User can increase the Hover time to the maximum allowed. 
- The value set by the user persists after leaving and returning to Settings. 

**Acceptance Criteria**:
All automation tests pass.

PR #586, opened 2022-04-21
[573] Hover Timing UI Tests

rowildcard:
https://github.com/willowtreeapps/vocable-ios/issues/573

# Description of Work
This a quick UI test to verify the Hover Timing buttons in Settings. 
1. Timing can be decreased to minimum & '-' button is disabled
2. Timing can be maxed & '+' button is disabled.
3. Custom settings persist after leaving and returning to Settings.

---

### DIFF
--- a/Vocable.xcodeproj/project.pbxproj
+++ b/Vocable.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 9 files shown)


rowildcard:
I'm closing this PR. These tests do not add as much value as some of the automation refactoring tickets. If those get done then we can implement the sensitivity and hover tests in unit tests instead.

PR #591, opened 2022-04-21
Adding Feature Flag for Listening Mode

jsmorgan42:
# Description of Work
This PR adds in a feature flag that by default turns off the listening mode feature. However, if the feature needs to be turned on, a toggle has been added to the listening mode debug menu (accessed by quadruple tapping the text view at the top in the main screen). 

---

### DIFF
--- a/Vocable/AppConfig.swift
+++ b/Vocable/AppConfig.swift
@@ -20,22 +20,22 @@ struct AppConfig {
         #endif
     }()
 
-    @PublishedDefault(key: "isHeadTrackingEnabled", defaultValue: AppConfig.isHeadTrackingSupported)
+    @PublishedDefault(key: .isHeadTrackingEnabled, defaultValue: AppConfig.isHeadTrackingSupported)
     static var isHeadTrackingEnabled: Bool
     static var isHeadTrackingSupported: Bool {
         return ARFaceTrackingConfiguration.isSupported
     }
 
-    @PublishedDefault(key: "dwellDuration", defaultValue: 1)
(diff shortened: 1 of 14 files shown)


Vocable/AppConfig.swift
```diff
@@ -20,22 +20,22 @@ struct AppConfig {
         #endif
     }()
 
-    @PublishedDefault(key: "isHeadTrackingEnabled", defaultValue: AppConfig.isHeadTrackingSupported)
+    @PublishedDefault(key: .isHeadTrackingEnabled, defaultValue: AppConfig.isHeadTrackingSupported)
```
Clstroud: Not sure that we need the key as an explicit label here, but we can refine separately

Clstroud (approved):
We paired on most of this so I had context. Good stuff! 🥇 💯

CananArikan:
After enabling the listen mode from debug menu, first question/phrase is written to the screen and I am getting a response for it. After the first question/phrase, when I continue to ask questions, application is listening as you can see from the Xcode console. But, what I am asking/talking is not written to the screen as an expected output text. 

Tested:
iPad Pro - Test Device - iOS 15.3.1
iPhone 11 - Simulator - iOS 15.4

https://user-images.githubusercontent.com/100222793/165151294-e0ecba33-d638-4f00-9f88-f76983b55583.mp4

jsmorgan42:
@CananArikan Fix has been pushed up!

CananArikan:
@jsmorgan42 works perfect! Thanks!

Clstroud (approved):

PR #596, opened 2022-04-22
[587] Allow preset injection from tests

### LINKED ISSUE #587: Allow preset data injection for testing purposes
This development ticket is to create the option when the app is launched of which preset json will be used in the app thus allowing testers to create test data that could be injected into the app. 

There are a couple of concerns in the UI automation when it comes to testing the preset data (categories and their associated phrases). The preset data used by the app is in a json file (/Vocable/Supporting Files/presets.json). This file, therefore, is the source of truth. 

- VocableUITests does not have immediate access to the json file.
- Hardcoding the data in VocableUITests is error prone, doesn't provide enough flexibility, and doesn't use the source of truth.

Access to the json data is possible from VocableUITests which would ensure tests are using the source of truth. But most of the tests are around verifying functionality, not the exact preset category and phrase strings. The strings can be anything as long as the file they're in is the same one used by the app and the automation tests. 

So why not _create_ a source of truth in VocableUITests, inject it into the app instead of the preset json file. Now UITests can create the exact test data needed for automation.  

[Parent ticket 590](https://github.com/willowtreeapps/vocable-ios/issues/590)

Clstroud:
Closes #587

Presets_Builder

Implements the above in a sample XCTestCase while also making the main app more safely handle missing preset categories


Vocable/Common/LaunchArguments.swift
```diff
@@ -8,21 +8,23 @@
...
-}
 
-fileprivate extension CommandLine {
+    let keys: [Key]
 
-    static var customArguments: [LaunchArguments.Key] {
-        arguments.compactMap(LaunchArguments.Key.init)
+    private init() {
+        keys = []
     }
```
jsmorgan42, line 29: I like it 💯

Vocable/Features/Root/FreeResponseTextEditorConfigurationProvider.swift
```diff
@@ -61,19 +61,24 @@ final class FreeResponseTextEditorConfigurationProvider: TextEditorConfiguration
             return
         }
 
+        let context = NSPersistentContainer.shared.viewContext
```
jsmorgan42, line 64: Should we just be using the background context that's passed in?

jsmorgan42 (approved):
👨‍🎤 🎸 🔥

Vocable/Features/Root/FreeResponseTextEditorConfigurationProvider.swift
Clstroud: If we're doing read-only operations, it's totally fine to use the viewContext. Especially since that's our source of truth for what we're showing in the UI 😄

VocableUITests/Tests/PresetsOverrideTestCase.swift
```diff
@@ -0,0 +1,43 @@
...
+//  PresetsOverrideTestCase.swift
+//  VocableUITests
+//
+//  Created by Chris Stroud on 4/22/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import XCTest
+
+class PresetsOverrideTestCase: XCTestCase {
```
rowildcard, line 11: This works like a charm after making a few changes to BaseTest.swift and this file. 
1. In `PresetsOverrideTestCase.swift` delete lines 13-16, and 33, 34.
2. Move lines 17-32 (the app.configure) into `BaseTest.swift` right after line 18. 
3. Comment out lines 16-18 in `BaseTest.swift`.

rowildcard (commented):
This works with a couple of changes. 
I want to review it with Rudy and Canan.

rowildcard:
As discussed, we would like the also have an `id` for the `Category` type.

rowildcard (approved):

Clstroud:
@rowildcard Added support for Category ID and fixed some merge conflicts after merging down develop

PR #601, opened 2022-04-26
Removing Reset App Settings As Debug Feature

jsmorgan42:
# Description of Work
The `Reset App Settings` option in the main setting screen was set as a debug feature, making it not visible on a production build. This PR removes it as a debug feature.

## Notes to Test (Optional)
This change should be noticeable by running on the app using `AppStore` scheme (where debug features are disabled/hidden).

---

### DIFF
--- a/Vocable/Features/Settings/SettingsViewController.swift
+++ b/Vocable/Features/Settings/SettingsViewController.swift
@@ -62,7 +62,7 @@ final class SettingsViewController: VocableCollectionViewController, MFMailCompo
         }
 
         var isFeatureEnabled: Bool {
-            let debugFeatures: [SettingsItem] = [.resetAppSettings, .pidTuner]
+            let debugFeatures: [SettingsItem] = [.pidTuner]
             if debugFeatures.contains(self) {
                 return AppConfig.showDebugOptions
             }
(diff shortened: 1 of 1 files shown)


Vocable/Features/Settings/SettingsViewController.swift
```diff
@@ -160,7 +160,7 @@ final class SettingsViewController: VocableCollectionViewController, MFMailCompo
                               .pidTuner].filter(\.isFeatureEnabled))
         snapshot.appendSections([.externalURL])
         snapshot.appendItems([.privacyPolicy,
-                              .contactDevs].filter(\.isFeatureEnabled))
+                              .contactDevs])
```
Clstroud: Curious: Why is this being removed if it's still valid on 160?

rowildcard:
Test verified.

rowildcard (approved):

Vocable/Features/Settings/SettingsViewController.swift
jsmorgan42: It didn't seem necessary since that section doesn't need to filter out any features, but in retrospect it probably makes sense to keep it in in case we want to add a debug feature to this section

jsmorgan42: It didn't seem necessary since that section doesn't need to filter out any features, but in retrospect it probably makes sense to keep it in in case we want to add a debug feature to this section

jsmorgan42: I added it back in

PR #603, opened 2022-04-26
[556] Update Listening Mode localized strings

### LINKED ISSUE #556: Voice UI: Update the content in the Listen Mode Setting category
**Acceptance Criteria**:

Update copy in the Listen Settings view as follows:

Listening Mode toggle
When listening mode is enabled, anyone can ask a question out loud and Vocable will offer a selection of responses. Currently, the app supports either/or questions and questions that can be answered with yes, no, or a number.

Hey, Vocable toggle
When this shortcut is enabled, anyone can say “Hey, Vocable” to prompt the app to navigate to the listening mode screen.

**Design**:

Needs to be created

Clstroud:
Closes #556 

* [Gently] overhauls the localization mechanism to appease the xliff export process and remove some code noise
    - `String(localized: )` should be preferred over `NSLocalizedString`
    - ^ that is iOS 15 API but it fits nicely with our objectives for iOS 14 so I'm stealing it and making a backward-compatible extension
* New strings should have their comments entered in Localizable.strings rather than in code
* Made the Listening Mode preferences screen use the `VocableListCell` API

### DIFF
--- a/Vocable/Extensions/String+Localization.swift
+++ b/Vocable/Extensions/String+Localization.swift
@@ -1,0 +1,19 @@
+//
+//  String+Localization.swift
+//  Vocable
+//
+//  Created by Chris Stroud on 4/26/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import Foundation
+
+// Allows the new iOS 15 syntax/API to work on earlier iOS versions.
+// Not guaranteed to have identical behavior, but has been tested enough
+// to give confidence that it should work well enough for our purposes.
+extension String {
+    @available(iOS, obsoleted: 15.0, message: "Use String(localized keyAndValue:, table:, bundle:, locale:, comment:) instead")
+    init(localized key: String, table: String? = nil, bundle: Bundle = .main, locale: Locale = .current, comment: StaticString? = nil) {
+        self = bundle.localizedString(forKey: key, value: nil, table: table)
+    }
+}
--- a/Vocable/Features/Settings/ListeningMode/ListeningModeViewController.swift
+++ b/Vocable/Features/Settings/ListeningMode/ListeningModeViewController.swift
@@ -12,18 +12,51 @@ import Combine
 
 final class ListeningModeViewController: VocableCollectionViewController {
 
+    private enum Section: Int {
+        case listeningMode
+        case hotword
+    }
+
     private enum ContentItem: Int {
         case listeningModeEnabled
         case hotWordEnabled
+
+        var title: String {
+            switch self {
+            case .listeningModeEnabled:
+                return String(localized: "settings.listening_mode.listening_mode_toggle_cell.title")
+            case .hotWordEnabled:
+                return String(localized: "settings.listening_mode.hot_word_toggle_cell.title")
+            }
+        }
+
+        var accessibilityID: String {
+            switch self {
+            case .listeningModeEnabled:
+                return "listening_mode_toggle"
+            case .hotWordEnabled:
+                return "hot_word_toggle"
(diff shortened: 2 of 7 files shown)


Vocable/Extensions/String+Localization.swift
```diff
@@ -0,0 +1,19 @@
...
+
+// Allows the new iOS 15 syntax/API to work on earlier iOS versions.
+// Not guaranteed to have identical behavior, but has been tested enough
+// to give confidence that it should work well enough for our purposes.
+extension String {
+    @available(iOS, obsoleted: 15.0, message: "Use String(localized keyAndValue:, table:, bundle:, locale:, comment:) instead")
+    init(localized key: String, table: String? = nil, bundle: Bundle = .main, locale: Locale = .current, comment: StaticString? = nil) {
+        self = bundle.localizedString(forKey: key, value: nil, table: table)
+    }
+}
```
jsmorgan42, line 19: I think you mentioned this change works better with Crowdin, right? Was curious why exactly that's the case

Vocable/Features/Settings/ListeningMode/ListeningModeViewController.swift
```diff
@@ -63,41 +95,84 @@ final class ListeningModeViewController: VocableCollectionViewController {
...
+            let reconfigurableItems = [.hotWordEnabled, .listeningModeEnabled].filter(snapshot.itemIdentifiers.contains)
+            snapshot.reconfigureItems(reconfigurableItems)
+        }
+
+        dataSource.apply(snapshot, animatingDifferences: animated) { [weak self] in
+            // Workaround for diffable datasource not auto-reconfiguring on iOS 14
+            if #unavailable(iOS 15) {
+                self?.updateVisibleCellConfigurations()
+            }
+        }
```
jsmorgan42, line 122: Interesting, could you explain why we don't want to use `dataSource.apply()` here?

Vocable/Features/Settings/ListeningMode/ListeningModeViewController.swift
```diff
@@ -118,7 +193,7 @@ This shortcut makes it fast to kick off a conversation by saying something like
         let group = NSCollectionLayoutGroup.horizontal(layoutSize: groupSize, subitem: item, count: columnCount)
         group.interItemSpacing = .fixed(8)
 
-        let footerSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0), heightDimension: .estimated(500))
+        let footerSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0), heightDimension: .estimated(1))
```
jsmorgan42, line 196: Are we changing the estimated height to 1 since the value is arbitrary anyways?

jsmorgan42 (approved):
LGTM 💸 Had a couple questions though

Vocable/Features/Settings/ListeningMode/ListeningModeViewController.swift
Clstroud: Yes, the footer was taking up the full height of the screen otherwise, so this has it compressing back down to the actual intrinsic content size of the view

Vocable/Extensions/String+Localization.swift
Clstroud: We should chat about it, but yes. More the xliff export process than Crowdin, but it also forces us to be more intentional about comments in `Localizable.strings` which will indirectly benefit Crowdin. Our inline code localized string comments weren't being used anyway since the values in `Localizable.strings` were overriding them.

PR #618, opened 2022-04-28
Make debug gesture harder to discover

Clstroud:
Now it’s a triple tap where you hold the last tap for 2 seconds


SashankPatel (approved):
It works - verified on iPhone 8 simulator.

PR #632, opened 2022-05-04
[599] Initial Listening Mode analytics support

Clstroud:
* Adds support for Mixpanel tokens via CI + environment variables
* Sets up super properties
* Performs single test event

### DIFF
--- a/Vocable/Analytics/Analytics.swift
+++ b/Vocable/Analytics/Analytics.swift
@@ -63,6 +63,7 @@ class Analytics {
 
         guard let token = Analytics.token else {
             print("No Mixpanel token found. Analytics will not be reported.")
+            Mixpanel.initialize(token: "", trackAutomaticEvents: false)
             return
         }
 
--- a/Vocable/Analytics/Analytics.swift
+++ b/Vocable/Analytics/Analytics.swift
@@ -1,0 +1,59 @@
+//
+//  Analytics.swift
+//  Vocable
+//
+//  Created by Chris Stroud on 5/3/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import Foundation
+import Mixpanel
+
+private struct EnvironmentTokens {
+
+    private static let infoPlistKeyStaging = "MIXPANEL_TOKEN_STAGING"
+    private static let infoPlistKeyProduction = "MIXPANEL_TOKEN_PRODUCTION"
+
+    let staging: String
+    let production: String
+
+    init?(infoDictionary dict: [String: Any]) {
+        guard
+            let staging = dict[EnvironmentTokens.infoPlistKeyStaging] as? String,
+            let production = dict[EnvironmentTokens.infoPlistKeyProduction] as? String
+        else {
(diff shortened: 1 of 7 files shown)


Vocable/Analytics/Analytics.swift
```diff
@@ -0,0 +1,102 @@
...
+        guard
+            let staging = dict[EnvironmentTokens.infoPlistKeyStaging] as? String,
+            let production = dict[EnvironmentTokens.infoPlistKeyProduction] as? String
+        else {
+            return nil
+        }
+        self.staging = staging
+        self.production = production
+    }
+}
```
moyerr, line 31: I'm surprised we don't already have an abstraction around getting values out of the `Info.plist` file. I don't think it needs to be for this PR, but I would love a more general-purpose version of this for getting any `Info.plist` value using strongly-typed keys.

Vocable/Analytics/Analytics.swift
```diff
@@ -0,0 +1,102 @@
...
+        register(superProperty: superProperty, using: publisher.eraseToAnyPublisher())
+    }
+
+    private func register(superProperty: String, using publisher: AnyPublisher<Bool, Never>) {
+        publisher
+            .receive(on: queue)
+            .sink { newValue in
+                Mixpanel.mainInstance().registerSuperProperties([superProperty: newValue])
+            }.store(in: &cancellables)
+    }
```
moyerr: Just a couple API design thoughts:

- These methods kinda wrap functionality on `MixpanelInstance` so I wonder if they'd be better suited for an extension on that type.
- I'm not entirely sure the type erasure is providing any value here. Was the goal to provide an abstraction that doesn't depend on the concrete `CurrentValueSubject` type? If that's the case, maybe generics can provide the reusable abstraction we're looking for? I could definitely see something like this coming in handy if we have to handle more values that might be published in different ways:

```Swift
extension MixpanelInstance {
    func registerSuperProperty<P: Publisher>(
        _ name: String,
        using publisher: P,
        on queue: DispatchQueue
    ) -> AnyCancellable where P.Output: MixpanelType, P.Failure == Never {
        publisher
            .receive(on: queue)
            .sink { [weak self] value in
                self?.registerSuperProperties([name: value])
            }
    }
}
```

Vocable/Analytics/Analytics.swift
```diff
@@ -0,0 +1,102 @@
...
+
+    // MARK: - Events
+
+    func appDidLaunch() {
+        Mixpanel.mainInstance().track(event: "App Open")
+    }
```
moyerr, line 93: I was going to put some thoughts and ideas about implementation here, but since this is just a test event, we can defer providing the real foundation of our analytics API to a future PR

moyerr (approved):
I think we'll probably want to be calling `MixpanelInstance.track(_:)` on our dedicated analytics `DispatchQueue` to ensure the calls are made on the same thread that the super properties are updated, but since this is just a single test event, we'll defer that to the next analytics PR.

PR #634, opened 2022-05-05
Task - Added descriptor text to localized english strings file.

SashankPatel:
# Description of Work
There is no ticket for this, there was a ticket to lay the initial foundation of converting the NSLocalizedString to a different String method.  This work expands on that.

- Added descriptors to the english localized strings

## Notes to Test (Optional)
There should be no functional change.
A spot check of the app to make sure strings appear in english for button text, alerts, headings etc.  
If there was an issue with the function then none of the strings would appear.  If there was an issue with copy and paste then the app would not build.

---


jsmorgan42 (changes requested):
You'll notice at the bottom of the Localizable file there's the empty state strings that don't have comments and aren't using the new `String(localized: )` function in their respective files (`ListeningFeedbackSubmitView` & `ListeningFeedbackSubmitView`). 

Could you modify those as well? Sorry that's my bad, should have updated those before I merged!

jsmorgan42 (approved):
Looks great! 💯 🥇

rhsalas3:
Tested and verified. The app builds fine and all strings w/in the app appear as expected; copy is as expected. 

Verified on iPhone 11 - iOS 15.4 (simulator) as well as an iPad Pro - iOS 15.3.1 (physical).

PR #642, opened 2022-05-06
[637] Defaulting Listening Mode Feature Flag On

### LINKED ISSUE #637: Vocable iOS - Default the Listen Feature to ON
**Acceptance Criteria**:
In Vocable 1.4 we added a feature flag to disable the Listen Feature.
1.4.1 we will be releasing the listen feature therefore the flag will need to default: true.

- NOTE: another option is to remove the flag, but until we get some use out the feature and see how it does in production, we should keep the feature flag.

jsmorgan42:
closes #637

# Description of Work
The listening mode feature flag is now defaulted to true so that listening mode should show up on a fresh app install.

## Notes to Test
- Since the feature flag setting is saved on device, you'll want to remove the app and re-install to test that the feature flag is on by default.

---

### DIFF
--- a/Vocable/Features/Voice/ListenModeDebugView.swift
+++ b/Vocable/Features/Voice/ListenModeDebugView.swift
@@ -83,7 +83,7 @@ struct ListenModeDebugView: View {
     @ObservedObject private var storage = ListenModeDebugStorage.shared
 
     @AppStorage(.listeningModeFeatureFlagEnabled)
-    var listeningModeFeatureFlagEnabled = false
+    var listeningModeFeatureFlagEnabled = true
 
     @ViewBuilder
     private var listView: some View {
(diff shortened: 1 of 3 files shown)


Vocable/Features/Voice/ListenModeDebugView.swift
```diff
@@ -83,7 +83,7 @@ struct ListenModeDebugView: View {
     @ObservedObject private var storage = ListenModeDebugStorage.shared
 
     @AppStorage(.listeningModeFeatureFlagEnabled)
-    var listeningModeFeatureFlagEnabled = false
+    var listeningModeFeatureFlagEnabled = true
```
moyerr: Will this be sufficient to turn the feature on for existing installations? This is the default value, so if `listeningModeFeatureFlagEnabled` is already persisted on a device, I'm not sure this default will be used.

rhsalas3:
Tested and verified. 

I also confirmed the latest update to this PR, as of 11:18a on 5/10 (Tuesday) with the commit to ensure that user defaults do not affect the settings from the debug menu. In other words, if you turn OFF listen mode from the debug menu then rebuild the app, listen mode remains ON instead of being off from the previous session.

Tested on:
iPhone X - iOS 14.2
iPhone 8 Plus - iOS 14.8.1
iPad Pro - 15.3.1

Clstroud (approved):
Nice and concise 💯

Issue #562, opened 2022-04-12
VocableUITests Refactor: Consolidate Screen Classes and Element Queries; Refactor Helper Functions


rhsalas3:
## Description:
There are helper functions declared in some of our screen classes that do not make sense for that screen, but would be better suited in one of our others. Furthermore, there are elements declared which can be shared from the BaseScreen so that we are not re-declaring the same element query in multiple classes. Update: now that we are refactoring the test code to override preset data to inject our own categories/phrases we should also refactor our helper methods to use the types defined in `PresetsBuilders.swift`. 

This ticket is for refactoring the existing UI test screen classes to consolidate the helper functions and elements so that the code is more readable, and DRY.

**Acceptance Criteria**:
- After changes made, all tests pass in the automation suite

rhsalas3:
Because we are not actively working on this at the moment, I am moving the ticket back to the 'Ready to Select' column.

PR #648, opened 2022-05-11
Preset Data Injection Refactor: Pagination Tests

### LINKED ISSUE #589: Refactor CategoryPhrasesPaginationTests to use injected preset data
This ticket is for refactoring the test functions in CategoryPhrasesPaginationTests to use injected preset data.

This work is part of the overall effort outlined in https://github.com/willowtreeapps/vocable-ios/issues/590 (the parent ticket to this one)

Injection data is hardcoded as type `Presets` in setup(). The ids are assigned to the cell element for category and phrase.

`func injectPresetData() -> Presets {
        return Presets {
            Category(id: "general_category", "General") {
                Phrase(id: "general_be_patient", "Please be patient")
                Phrase(id: "general_donde_estoy", languageCode: "es", "No sé donde estoy")
            }`

**Acceptance Criteria**:

All tests in SettingsScreenTests pass

rhsalas3:
closes #589 

### Description of Work
Refactored all of the pagination tests to leverage the preset data injection during app startup. We were able to create 3 categories that worked for all tests between the CategoryPhrasesPaginationTests and MainScreenPaginationTests classes. So we created a new base test class, PaginationBaseTest, which then sets up the app with the injected presets.

---

### Notes to Test:
On this branch only the MainScreenPaginationTests, CategoryPhrasesPaginationTests, SettingsScreenTests, and MainScreenTests should run. During this effort we re-enabled SettingsScreenTests, which did not need to be refactored, but they should run and pass along with the pagination tests.

### DIFF
--- a/VocableUITests/Screens/MainScreen.swift
+++ b/VocableUITests/Screens/MainScreen.swift
@@ -73,4 +73,10 @@ class MainScreen: BaseScreen {
         return app.cells.staticTexts.containing(predicate).element
     }
     
+    static func navigateToSettingsAndOpenCategory(name: String) {
+        MainScreen.settingsButton.tap()
+        SettingsScreen.categoriesButton.tap()
+        SettingsScreen.openCategorySettings(category: name)
+    }
+    
 }
--- a/VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
+++ b/VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
@@ -8,18 +8,12 @@
 
 import XCTest
 
-class CategoryPhrasesPaginationTests: CustomPhraseBaseTest {
+class CategoryPhrasesPaginationTests: PaginationBaseTest {
     
-    /* TODO: There is an existing bug(s) that causes this test to fail with
-     animations turned off. We will re-enable this test once they're fixed.
-     https://github.com/willowtreeapps/vocable-ios/issues/597
-     https://github.com/willowtreeapps/vocable-ios/issues/594
-     */
     func testCanNavigatePages() {
-        // Add enough phrases to have 2 pages.
-        listOfPhrases.forEach { phrase in
-            CustomCategoriesScreen.addPhrase(phrase)
-        }
+        // Navigate to our test category
+        MainScreen.navigateToSettingsAndOpenCategory(name: categoryTwo.presetCategory.utterance)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
         
         // Verify that the user is on the first page and the next page buttons are enabled.
         VTAssertPaginationEquals(1, of: 2, enabledArrows: .both)
@@ -36,36 +30,29 @@ class CategoryPhrasesPaginationTests: CustomPhraseBaseTest {
             CustomCategoriesScreen.paginationLeftButton.tap()
             XCTAssertEqual(CustomCategoriesScreen.currentPageNumber, pageNumber)
         }
-        
     }
     
     func testPagesAdjustToNewPhrases() {
-        // Add enough phrases (7) to fill one page.
-        for phrase in listOfPhrases.startIndex..<8 {
-            CustomCategoriesScreen.addPhrase(listOfPhrases[phrase])
-        }
-        
         // Verify that the user is on the first page.
-        XCTAssertEqual(CustomCategoriesScreen.currentPageNumber, 1)
+        MainScreen.navigateToSettingsAndOpenCategory(name: categoryOne.presetCategory.utterance)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
         
         // Add 1 more phrase to push the total number of pages to 2.
-        let originalPageCount = CustomCategoriesScreen.totalPageCount
-        let expectedPageCountAfterAddingPhrase = originalPageCount + 1
         CustomCategoriesScreen.addRandomPhrases(numberOfPhrases: 1)
-        XCTAssertEqual(CustomCategoriesScreen.totalPageCount, expectedPageCountAfterAddingPhrase)
+        VTAssertPaginationEquals(1, of: 2, enabledArrows: .both)
         
         // Remove the additional phrase to verify that the page count reduces, back to the original count
         CustomCategoriesScreen.categoriesPageDeletePhraseButton.firstMatch.tap()
         SettingsScreen.alertDeleteButton.tap(afterWaitingForExistenceWithTimeout: 0.25)
-        XCTAssertEqual(CustomCategoriesScreen.totalPageCount, originalPageCount)
+        VTAssertPaginationEquals(1, of: 1, enabledArrows: .none)
     }
     
     // It is expected that the pagination left and right arrows are disabled when there is only 1 total page
     func testNextPageButtonsDisabled() {
-        // Add 1 phrase so that the pagination buttons appear
-        XCTAssertTrue(CustomCategoriesScreen.emptyStateAddPhraseButton.exists, "Expected a new category to be in empty state.")
-        CustomCategoriesScreen.addRandomPhrases(numberOfPhrases: 1)
-    
+        // Navigate to our test category and open the 'Edit Phrases' screen
+        MainScreen.navigateToSettingsAndOpenCategory(name: categoryOne.presetCategory.utterance)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
+        
         // Verify the page counts and that buttons appear; both buttons are disabled.
         VTAssertPaginationEquals(1, of: 1, enabledArrows: .none)
     }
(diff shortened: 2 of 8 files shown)


VocableUITests/Screens/MainScreen.swift
```diff
@@ -73,4 +73,10 @@ class MainScreen: BaseScreen {
         return app.cells.staticTexts.containing(predicate).element
     }
     
+    static func navigateToSettingsAndOpenCategory(name: String) {
```
Clstroud, line 76: Now that we have an actual `Category` type that encapsulates name/ID/etc, would it perhaps be cleaner to have that type be your primary unit of currency for these types of functions instead of passing around strings?

VocableUITests/Tests/PaginationBaseTest.swift
```diff
@@ -0,0 +1,61 @@
...
+//  CustomCategoriesBaseTest.swift
+//  VocableUITests
+//
+//  Created by Rudy Salas and Canan Arikan on 5/06/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import XCTest
+
+class PaginationBaseTest: XCTestCase {
```
Clstroud, line 11: I don't think inheritance is the _wrong_ choice here, but I _do_ think you could probably solve this with a protocol [extension] instead so as to keep all your `XCTestCase` instances consistent with how they set up their data.

Such a protocol could provide the preset data, and every entity that conforms to the protocol would have access to it.

VocableUITests/Tests/PaginationBaseTest.swift
```diff
@@ -0,0 +1,61 @@
...
+    
+    let categoryOne = Category(id: "first_category", "Category 1") {
+        Phrase("Please help")
+        Phrase("Hello")
+        Phrase("I need a blanket")
+        Phrase("I need a blanket")
+        Phrase("I need a blanket")
+        Phrase("I need a blanket")
+        Phrase("I need a blanket")
+        Phrase("I need a blanket")
```
Clstroud: Two thoughts:
1. These all have the same contents, so maybe this array of phrases can be privately defined once as a computed property and then each of the categories can reference it?
2. Would it be convenient to have a means of making `n` copies of a particular Phrase? e.g. `Phrases(repeating: "I need a blanket", count: 6)`? Not positive it'll work because of how Result Builders are, but it might!

Clstroud (commented):
Have just a few thoughts before approving anything. Happy to chat about any them. Looking good overall so far 🙂

VocableUITests/Tests/CategoryPhrasesPaginationTests.swift
```diff
@@ -36,36 +30,29 @@ class CategoryPhrasesPaginationTests: CustomPhraseBaseTest {
...
-        XCTAssertEqual(CustomCategoriesScreen.currentPageNumber, 1)
+        MainScreen.navigateToSettingsAndOpenCategory(name: categoryOne.presetCategory.utterance)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
         
         // Add 1 more phrase to push the total number of pages to 2.
-        let originalPageCount = CustomCategoriesScreen.totalPageCount
-        let expectedPageCountAfterAddingPhrase = originalPageCount + 1
         CustomCategoriesScreen.addRandomPhrases(numberOfPhrases: 1)
-        XCTAssertEqual(CustomCategoriesScreen.totalPageCount, expectedPageCountAfterAddingPhrase)
+        VTAssertPaginationEquals(1, of: 2, enabledArrows: .both)
```
rowildcard, line 42: Nice!

VocableUITests/Screens/MainScreen.swift
rhsalas3: Yes, we were thinking the same thing. When we went down that route, the number of changes grew to the point where we were not comfortable doing it on this PR for its size. We wanted to keep each PR to a minimum. 

We decided to make the effort to refactor these types of functions when we revisit our intended work on #562. 

I have made sure the description of work on that ticket reflects our intention to make these changes.

VocableUITests/Tests/PaginationBaseTest.swift
rhsalas3: I remember our conversation about using a protocol to represent an app screen instead of inheriting each screen from BaseScreen. With the test classes I'm not sure we want the setup to be consistent among them, as we want each setup to be specific to the tests. As we worked through this refactor we originally had a setup in each of the pagination test classes. But we saw where the setup could be shared, leading to this base test. In some of the other test classes we don't override the presets at all.

🤔  I may be missing your point. So let's chat more about this. 🤔

VocableUITests/Tests/PaginationBaseTest.swift
rhsalas3: In these tests, what the phrases are is arbitrary and insignificant to the tests. With that said, I like the idea of defining a private computed property. Looks and reads better. Plus, it's probably overkill for making `Phrases(repeating: ...)` since 99% of the other tests will just need a category with one phrase in it.

I will implement the computed property.

rowildcard (commented):
Test verified locally.

VocableUITests/Tests/PaginationBaseTest.swift
```diff
@@ -0,0 +1,61 @@
...
+        Phrase("I need a blanket")
+    }
```
jsmorgan42: Someone get this poor guy a blanket

VocableUITests/Tests/MainScreenPaginationTests.swift
```diff
@@ -8,73 +8,40 @@
...
         VTAssertPaginationEquals(1, of: 2)
         
         // Delete one of the phrases to reduce the total number of pages to 1.
-        SettingsScreen.navigateToSettingsCategoryScreen()
-        SettingsScreen.openCategorySettings(category: customCategoryName)
+        MainScreen.navigateToSettingsAndOpenCategory(name: categoryOne.presetCategory.utterance)
         CustomCategoriesScreen.editCategoryPhrasesButton.tap()
         CustomCategoriesScreen.categoriesPageDeletePhraseButton.firstMatch.tap()
-        SettingsScreen.alertDeleteButton.tap()
+        SettingsScreen.alertDeleteButton.tap(afterWaitingForExistenceWithTimeout: 0.5)
```
jsmorgan42, line 21: Is this to make a tests less flaky? If so, we may also want to add this timeout wait in the `testDeleteCustomPhrase()` function in `CustomPhraseTests`

VocableUITests/Tests/PaginationBaseTest.swift
```diff
@@ -0,0 +1,61 @@
...
+    
+    let categoryThree = Category(id: "third_category", "Category 3") {
```
jsmorgan42: I wonder if these category names could be a little more descriptive, maybe they correlate to the number of phrases? Something like `sevenPhrasesCategory`. Just a thought, feel free to ignore it if you think it would be more confusing

VocableUITests/Tests/PaginationBaseTest.swift
rhsalas3: Per our huddle session, we implemented a private function within the PaginationBaseTest to build an array of the same Phrase given a count.

🦖

VocableUITests/Tests/PaginationBaseTest.swift
rhsalas3: Discussed in a Slack Huddle. We will revisit the idea of using a protocol instead of base test classes. This work is tracked in ticket #654

VocableUITests/Screens/MainScreen.swift
rhsalas3: Discussed in a Slack Huddle. We will refactor the helper functions separate from this PR to reduce the size of each PR.

VocableUITests/Tests/PaginationBaseTest.swift
rhsalas3: @Clstroud  Please take a look at #654 when you have a chance and make sure that I've captured the intent and needs for implementing a protocol for our UI tests.

VocableUITests/Tests/MainScreenPaginationTests.swift
rhsalas3: Actually I forget why we needed that. The test wasn't flaky here, it just had something to do with turning off animations. 

Long story short, it doesn't hurt to add to `testDeleteCustomPhrase()` and it may be needed anyway when we get to refactoring `CustomPhraseTests`. 

Change made! 😉

VocableUITests/Tests/PaginationBaseTest.swift
rhsalas3: Always a good thing to have descriptive names. 

I will implement this change.

jsmorgan42 (approved):
Love how much simpler this makes the UI tests! 🥇

VocableUITests/Tests/PaginationBaseTest.swift
Clstroud: Sounds good enough to me. Thanks for making that!

Clstroud (approved):
Nice job @rhsalas3 🏆

PR #649, opened 2022-05-11
[613] General Analytics

### LINKED ISSUE #613: Mixpanel Integration: General Analytics
The purpose of this ticket is to implement Mixpanel Integration so that we can track events related to general usage.

**Acceptance Criteria:** 
- All events tracked are not tied to user private information
- Events track meets apple requirements to maintain [App Store Privacy disclosure](https://developer.apple.com/app-store/app-privacy-details/#linked-data) of "Data Not Linked to You"

**Events captured includes:**
1. Preset Category/Phrase Selection
    - When a preset category is selected
    - When a preset phrase is selected from a preset category 
3. Keyboard from main page
    - When keyboard opens 
    - When the speak option is selected
    - When favorite is selected
4. Timing and Sensitivity Tracking
     - Hover time changed 
     - Cursor Sensitivity Changed 
5. Usage events
     - When application is opened
     - When application is backgrounded
     - When application is is closed
6. Device Orientation
     - Super property for each event showing device orientation

     
Links:
- [Tracking Plan Provided by Analytics Team]- See WT Confluence

jsmorgan42:
closes #613

# Description of Work
Adding the remaining analytics calls including two new super properties for hover time & cursor sensitivity. 

---

### DIFF
--- a/Vocable/Analytics/Analytics.swift
+++ b/Vocable/Analytics/Analytics.swift
@@ -89,19 +89,9 @@ class Analytics {
     private func registerSuperProperties() {
         cancellables = [mixPanel.registerSuperProperty("Listening Mode Enabled", using: listeningMode.$listeningModeEnabledPreference, on: queue),
                         mixPanel.registerSuperProperty("'Hey Vocable' Enabled", using: listeningMode.$hotwordEnabledPreference, on: queue),
-                        mixPanel.registerSuperProperty("Head Tracking Enabled", using: AppConfig.$isHeadTrackingEnabled, on: queue)]
-
-        AppConfig.$cursorSensitivity
-            .receive(on: queue)
-            .sink { [weak self] sensitivity in
-                self?.mixPanel.registerSuperProperties(["Cursor Sensitivity": sensitivity.analyticsDescription])
-            }.store(in: &cancellables)
-
-        AppConfig.$selectionHoldDuration
-            .receive(on: queue)
-            .sink { [weak self] hoverTime in
-                self?.mixPanel.registerSuperProperties(["Hover Time": hoverTime.analyticsDescription])
-            }.store(in: &cancellables)
+                        mixPanel.registerSuperProperty("Head Tracking Enabled", using: AppConfig.$isHeadTrackingEnabled, on: queue),
+                        mixPanel.registerSuperProperty("Cursor Sensitivity", using: AppConfig.$cursorSensitivity.map(\.analyticsDescription), on: queue),
+                        mixPanel.registerSuperProperty("Hover Time", using: AppConfig.$selectionHoldDuration.map(\.analyticsDescription), on: queue)]
     }
 
     // MARK: - Events
--- a/Vocable/Analytics/Analytics.swift
+++ b/Vocable/Analytics/Analytics.swift
@@ -64,7 +64,7 @@ class Analytics {
     private var cancellables = Set<AnyCancellable>()
     private let queue = DispatchQueue(label: "Analytics-Processing")
 
-    private let mixPanel: MixpanelInstance
+    let mixPanel: MixpanelInstance
 
     private let listeningMode = AppConfig.listeningMode
 
@@ -134,7 +134,6 @@ extension Analytics.Event {
 
     static let appOpened = Self(name: "App Opened")
     static let appBackgrounded = Self(name: "App Backgrounded")
-    static let appClosed = Self(name: "App Closed")
 
     // MARK: Keyboard Phrase
 
@@ -143,6 +142,10 @@ extension Analytics.Event {
     static let phraseFavorited = Self(name: "Keyboard Phrase Favorited")
 
     // MARK: Categories
+
+    static func categorySelected(_ category: Category) -> Self {
+        Self(name: "Category Selected", properties: ["Category": category.analyticsName])
+    }
 
     static let newCategoryCreated = Self(name: "New Category Created")
 
@@ -166,8 +169,8 @@ extension Analytics.Event {
         Self(name: "Preset Phrase Edited", properties: ["Phrase": utterance])
     }
 
-    static func presetPhraseSelected(_ utterance: String) -> Self {
-        Self(name: "Preset Phrase Selected", properties: ["Phrase": utterance])
+    static func phraseSelected(_ phrase: Phrase) -> Self {
+        Self(name: "Phrase Selected", properties: ["Phrase": phrase.analyticsName])
     }
 
     // MARK: Listening Mode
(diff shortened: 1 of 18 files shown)


Vocable/Analytics/Analytics.swift
```diff
@@ -90,6 +90,18 @@ class Analytics {
...
+            .receive(on: queue)
+            .sink { [weak self] sensitivity in
+                self?.mixPanel.registerSuperProperties(["Cursor Sensitivity": sensitivity.description])
+            }.store(in: &cancellables)
+
+        AppConfig.$selectionHoldDuration
+            .receive(on: queue)
+            .sink { [weak self] hoverTime in
+                self?.mixPanel.registerSuperProperties(["Hover Time": "\(hoverTime)s"])
+            }.store(in: &cancellables)
```
jsmorgan42: I'm sure there's a better way to do this I just couldn't think of one in the moment

moyerr: @jsmorgan42 Isn't this what our extension on `MixpanelInstance` was for

jsmorgan42: We can't use that extension because we need to customize how the value is being displayed. The other ones are just boolean values so they could use that same function.

Vocable/Features/Settings/TimingSensitivity/Model/CursorSensitivity.swift
```diff
@@ -22,7 +27,14 @@ enum CursorSensitivity: Int, Codable {
...
+    var description: String {
+        switch self {
+        case .low:
+            return "low"
+        case .medium:
+            return "medium"
+        case .high:
+            return "high"
+        }
+    }
```
Clstroud: I can see an argument either way here, but a small part of me feels weird about relying on `description` to remain stable over time for analytics consistency. I almost think a `CustomAnalyticsReportable` or similar protocol that provides a dedicated means for specifying the analytics value would make sense. It also provides a natural means to format the string, as you're currently doing by hand with hover time.

Vocable/Features/Settings/TimingSensitivity/Model/CursorSensitivity.swift
```diff
@@ -9,6 +9,11 @@
 import UIKit
 
 enum CursorSensitivity: Int, Codable {
+
```
Clstroud:
```suggestion
```

Vocable/Features/Root/CategoryDetailViewController.swift
```diff
@@ -188,6 +188,10 @@ class CategoryDetailViewController: PagingCarouselViewController, NSFetchedResul
 
                 self.lastUtterance = utterance
 
+                if !phrase.isUserGenerated, !phrase.isUserRenamed {
```
Clstroud: Maybe an extension on both `Phrase` and `Category` that says `var isAnalyticsReportable: Bool` or something would cut down on the "stop-and-think-about-it" logic in all these related if-statements?

Vocable/Analytics/Analytics.swift
```diff
@@ -128,41 +140,65 @@ extension Analytics.Event {
...
+
+    // MARK: Settings
+
+    static let hoverTimeChanged = Self(name: "Hover Time Settings Changed")
+    static let cursorSensitivityChanged = Self(name: "Cursor Sensitivity Changed")
+
+    static let listeningModeChanged = Self(name: "Listening Mode Settings Changed")
+    static let heyVocableModeChanged = Self(name: "'Hey Vocable' Settings Changed")
+
+    static let headingTrackingChanged = Self(name: "Head Tracking Settings Changed")
```
Clstroud, line 193: More of a thought to consider than anything (feel free to ignore if you don't like it): these might feel more "Cocoa" if they were `...DidChange` rather than `Changed`

Clstroud (commented):
A few comments before I approve anything. Liking what you've got so far 🥇

Vocable/Features/Root/CategoryDetailViewController.swift
jsmorgan42: I also thought about doing something like that, I think that's a good idea

Vocable/Analytics/Analytics.swift
jsmorgan42: Fair enough, I'll make that change

jsmorgan42: Actually on second thought, I think I'd prefer these functions to match the same naming convention that we have in the spec sheet

Vocable/Features/Settings/TimingSensitivity/Model/CursorSensitivity.swift
jsmorgan42: I think that idea, I'll look into it

Vocable/Analytics/Analytics.swift
moyerr: @jsmorgan42  That's the beauty of Combine:

```Swift
mixPanel.registerSuperProperty(
  "Cursor Sensitivity",
  using: AppConfig.$cursorSensitivity.map(\.description),
  on: queue
)
```

jsmorgan42: Ah right, that makes sense 🤦

Clstroud (approved):
Nicely done 🏆

CananArikan:
Hey @Ly-Kyli, @jsmorgan42, @SashankPatel,

@rowildcard, @rhsalas3, and I have some questions below.

1. Preset Category/Phrase Selection

- We don't have implementation of selecting a preset category, there is one in the Mixpanel Tracking Plan sheet. Do we want to implement it?
- Selection method is not appearing for our events in the Mixpanel. Is it still something we intend to include?

2. Keyboard from main page

- We track when favorite is selected, do we want to track when it is deselected?

3. Usage events

- When application is closed, we don't have this implementation and also nothing specified in the Mixpanel Tracking Plan sheet. Do we still want an event for this?
- We have implementation of App Backgrounded, but it is not specified in the sheet. Should we include it in the sheet?

CananArikan:
Preset categories are ordered as General, Basic Needs, Personal Care, Conversation, Environment, 123, My Sayings, Recents, Listen. When iPhone is on the portrait mode, we have one category as selected in the main screen. If we switch to Listen by clicking right arrow, all categories between General and Listen are tracked as `Category Selected` in Mixpanel, which is too much useless data. As @jsmorgan42, @kyliefk, @rhsalas3, @rowildcard, and I discussed, we need an implementation to add a `Category` property to the `Phrase Selected` event. In this case, we will track the phrases with category information for more meaningful reports.

CananArikan:
Thanks for the updates! Test verified!

Devices: 
- iPad Pro - Test Device - iOS 15.3.1 
- iPhone 11 - Simulator - iOS 15.4

Version: 
- 1.4.1-1

PR #650, opened 2022-05-11
[563] Remove debug localizations

### LINKED ISSUE #563: Vocable iOS: Remove Debug text localization.
**Describe the bug**
Currently debug menu empty state text is in the localization files.  These should be removed and not be localized.
Also there is a string that is in the German Localization there is a %1$@. %2$@ string, this should not be there.

**To Reproduce**
- Quadruple click in the text field to bring up the debug menu.
- If there are entries scroll to the bottom of the list and tap on the clear all button and confirm.

**Actual behavior**
Empty state appears with the following strings: 
"No Entries"
"The most recent listening sessions will be recorded here for easy debugging"

These strings appear in Crowdin to be localized but should not be.

Clstroud:
Closes #563 

# Description of Work
Removes debug-related localizations from the project

## Notes to Test
* I'm curious how the xliff export process will react to this after we merge into develop. I am not certain how these strings got here in the first place, so I'm crossing my fingers that there isn't something funky going on with the xliff exporter and SwiftUI that might cause them to pop back up. I don't think that's likely, since there are other Text() instances with string literals that were not localized in kind residing in the same file. Odds are it was briefly manually translated and then removed from the development localization but not the others.

### DIFF
--- a/Vocable/Supporting Files/ar.lproj/Localizable.strings
+++ b/Vocable/Supporting Files/ar.lproj/Localizable.strings
@@ -28,9 +28,6 @@
 /* Toast message that appears after saving edits made to the Category title */
 "category_editor.toast.successfully_saved.title" = "تم حفظ التغييرات";
 
-/* This seems to not be used currently */
-"debug.assertion.presets_file_not_found" = "لم يتم العثور على إعدادات مسبقة";
-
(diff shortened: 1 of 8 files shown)


SashankPatel (approved):
Lets see what happens.

PR #653, opened 2022-05-11
#598 - Fix Listening Mode bug 🗣️ 🐞 

### LINKED ISSUE #598: Voice Recognition not always triggering
**Describe the bug**
When navigating between categories if you navigate to the listen category very quickly the Voice recognition doesn't seem to activate.  
Listen feature seems to be working fine when saying "Hey vocable" the listening is being picked up so it may seem like its only when navigating to the category.

**To Reproduce**
Quickly toggle between the listen category and a non-listen category i.e. General.

**Expected behavior**
When navigating to the Listen category I should be able to talk to vocable and the app should pick up what I am saying.

**Actual behavior**
Sometimes the app is not picking up what I am saying.
Sometime the Listen will pick up the initial phrase but will not pick up subsequent phrases.

**Device Information**
Tested on Iphone 8 simulator as well as a physical iPad device (7th or 9th gen).

moyerr:
Closes #598

Listening mode would sometimes stop working if the user quickly switched between the Listening category and another category. This was caused by the fact that, when switching back, a new `ListeningResponseViewController` was initialized before the old one was deinitialized. So the `SpeechRecognitionController` was still receiving signals from the "old" insance, which were interfering with the signals from the "new" one.

This was resolved by adding an activate/deactivate mechanism to the `ListeningResponseViewController`, in which it can setup and tear down its relationship with the `SpeechRecognitionController`.

### DIFF
--- a/Vocable/Features/Root/RootViewController.swift
+++ b/Vocable/Features/Root/RootViewController.swift
@@ -96,7 +96,6 @@ import SwiftUI
         } else if #available(iOS 14.0, *), destinationIsListeningMode {
             let vc = ListeningResponseViewController()
             vc.delegate = self
-            vc.activate(withSpeechRecognizer: .shared)
             utterancePublisher = vc.$lastUtterance
             viewController = vc
         } else {
@@ -114,6 +113,10 @@ import SwiftUI
                     updateOutputLabelText(nil, isDictated: false)
                 }
             }
+
+            SpeechRecognitionController.shared.stopTranscribing()
+        } else {
+            SpeechRecognitionController.shared.startTranscribing()
         }
 
         utteranceCancellable = utterancePublisher.receive(on: DispatchQueue.main)
@@ -129,13 +132,6 @@ import SwiftUI
         let childrenToDisposeOf = children.filter {
             ![categoryCarousel, viewController].contains($0)
         }
-
-        // https://github.com/willowtreeapps/vocable-ios/issues/598
-        // Deactivate any living instances of ListeningResponseViewController
-        // so they will immediately stop attempting to interact with the SpeechRecognitionController
-        childrenToDisposeOf
-            .compactMap { $0 as? ListeningResponseViewController }
-            .forEach { $0.deactivate() }
 
         let contentTransform = CGAffineTransform.identity
         let transitionTransform = CGAffineTransform(translationX: 0,
--- a/Vocable/Features/Root/RootViewController.swift
+++ b/Vocable/Features/Root/RootViewController.swift
@@ -96,6 +96,7 @@ import SwiftUI
         } else if #available(iOS 14.0, *), destinationIsListeningMode {
             let vc = ListeningResponseViewController()
             vc.delegate = self
+            vc.activate(withSpeechRecognizer: .shared)
             utterancePublisher = vc.$lastUtterance
             viewController = vc
         } else {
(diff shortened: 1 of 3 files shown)


Vocable/Features/Voice/ListeningResponseViewController.swift
```diff
@@ -285,6 +285,29 @@ final class ListeningResponseViewController: VocableViewController {
...
+
+        transcriptionCancellable?.cancel()
+        availabilityCancellable?.cancel()
+        permissionsCancellable?.cancel()
+        classificationCancellable?.cancel()
+
+        transcriptionCancellable = nil
+        availabilityCancellable = nil
+        permissionsCancellable = nil
+        classificationCancellable = nil
```
jsmorgan42: Is it necessary to also make them nil? Wouldn't just calling `cancel()` suffice?

moyerr: The main motivation for this was the fact that there are a few places where there are nil checks for the cancellables. So basically, I didn't want us to be in a state where they are cancelled but not nil, because then we might skip over some logic when we actually don't intend to.

jsmorgan42 (approved):
I like it! 💯

Vocable/Features/Voice/SpeechRecognitionController.swift
```diff
@@ -264,6 +264,7 @@ class SpeechRecognitionController: NSObject, SFSpeechRecognitionTaskDelegate, SF
         guard isPaused else {
             return
         }
+        print("RESUME LISTENING...")
```
Clstroud, line 267: Tangential: Any thoughts on integrating a logging framework as a future ticket?

Vocable/Features/Root/RootViewController.swift
```diff
@@ -129,6 +130,13 @@ import SwiftUI
             ![categoryCarousel, viewController].contains($0)
         }
 
+        // https://github.com/willowtreeapps/vocable-ios/issues/598
+        // Deactivate any living instances of ListeningResponseViewController
+        // so the will immediately stop attempting to interact with the SpeechRecognitionController
```
Clstroud:
```suggestion
        // so they will immediately stop attempting to interact with the SpeechRecognitionController
```

Vocable/Features/Voice/SpeechRecognitionController.swift
moyerr: @Clstroud I've thought about it many times! I can create a ticket for the ol' backlog

Vocable/Features/Voice/ListeningResponseViewController.swift
```diff
@@ -285,6 +285,29 @@ final class ListeningResponseViewController: VocableViewController {
...
+    /// Activates this instance with a Speech Recognizer
+    ///
+    /// - Parameter speechRecognizer: The `SpeechRecognitionController` for transcriptions
+    func activate(withSpeechRecognizer speechRecognizer: SpeechRecognitionController) {
+        speechRecognizerController = speechRecognizer
+    }
+
+    /// Deactivates this instance by removing the speech recognizer and cancelling
+    /// any active subscriptions to the speech recognizer.
+    func deactivate() {
```
Clstroud: Not arguing this strongly, but at the call site I interpreted this as doing something more than it actually is. "Activate" struck me as "start doing audio stuff." which we are leaving to the discretion of the child view controller.

Makes me wonder if having `deactivate` get called from the `didSet` of the `speechRecognitionController` might be preferable?

Vocable/Features/Root/RootViewController.swift
```diff
@@ -158,9 +166,15 @@ import SwiftUI
...
         func finalize(_ didFinish: Bool) {
             for inactiveViewController in childrenToDisposeOf {
-                inactiveViewController.removeFromParent()
+                guard inactiveViewController.parent != nil else {
+                    continue
+                }
+
+                inactiveViewController.willMove(toParent: nil)
                 inactiveViewController.view.removeFromSuperview()
+                inactiveViewController.removeFromParent()
```
Clstroud, line 171: >If you are implementing your own container view controller, it must call the willMove(toParent:) method of the child view controller before calling the [removeFromParent()](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621425-removefromparent) method, passing in a parent value of nil.

lol. No clue who wrote this 👀  Good catch.

Clstroud (approved):
Couple of thoughts, but looks good! 💯

Clstroud (approved):
Nice and concise 💯

CananArikan:
Test verified.
iPad Pro - Test Device - iOS 15.3.1
iPhone 11 - Simulator - iOS 15.4

PR #659, opened 2022-05-13
#646 | Treat Listening Mode like a hidden category when turned off

### LINKED ISSUE #646: Listen Category missing from Settings > Categories list when turned off
### Description: 
There are 2 flows for turning off listening mode:
1. Settings > 'Listening Mode' setting; OR 
2. Settings > 'Categories and Phrases' > edit the 'Listen' category.

When using flow `2` the category remains in the list and behaves as any other category would when hidden. However, when using flow `1` the category is removed from the list of categories. The 'Listen' category should remain in the list of categories even when turned off from within the Settings > 'Listening Mode' setting.

Update: currently it is expected behavior for flow `2`. But I think the team should rethink the flow given that toggling the listen mode behaves differently on this screen, than when editing other categories. Per the comment discussion on this issue, it seems to make the most sense to just keep flow `1` and not even list the 'Listen' category from w/in the Categories list screen.

**Steps to Reproduce:**
1. From the main screen, navigate to Settings > 'Listening Mode'. 
2. Tap on the 'Listening Mode' toggle button to turn the feature OFF.
3. Return to Settings
4. Tap 'Categories and Phrases'
5. Note that the 'Listen' category is not in the list.

**Expected behavior:**
After turning off Listen Mode from within Settings > 'Listen Mode', a user should still be able to re-enable listening mode from Settings > Categories and Phrases > edit 'Listen' category.

**Actual behavior:**
After turning off Listen Mode from within Settings > 'Listen Mode' the 'Listen' category is removed from the list of categories. 

-----

**Device Information:**
Latest build on develop as of 05.10.2022

moyerr:
Closes #646

# Description of Work

This PR changes the way we treat the Listening Mode category when displaying all categories in the Edit Categories screen. Previously, Listening mode would be completely removed from the list when disabled, however, we are now treating it like a hidden category:

- It gets placed after the enabled categories
- Its reorder buttons become disabled
- It can still be tapped on by the user to re-enable Listening Mode

### DIFF
--- a/Vocable/Extensions/Category+Helpers.swift
+++ b/Vocable/Extensions/Category+Helpers.swift
@@ -86,12 +86,15 @@ extension Category {
 
     static func visibleCategoriesPredicate(includeUserHidden: Bool = true) -> NSPredicate {
         var predicate = !Predicate(\Category.isUserRemoved)
+
         if !includeUserHidden {
             predicate &= !Predicate(\Category.isHidden)
+
+            if !AppConfig.listeningMode.isEnabled {
+                predicate &= !Predicate(\Category.identifier, equalTo: Category.Identifier.listeningMode)
+            }
         }
-        if !AppConfig.listeningMode.isEnabled {
-            predicate &= !Predicate(\Category.identifier, equalTo: Category.Identifier.listeningMode)
-        }
+
         return predicate
     }
     


jsmorgan42 (approved):
💥

rhsalas3:
Tested and works like a charm 🌟  🤩  🌟 

Tested on iPhone 8 Plus (iOS 14.8.1) and an iPhone X (iOS 14.2)

Clstroud (approved):
Nice 🚢

PR #661, opened 2022-05-13
[563] Ensure SwiftUI debugging strings are not localized

Clstroud:
An addendum to #563 

Ensures unwanted debug UI strings are not erroneously localized.


moyerr (approved):
🔠

PR #663, opened 2022-05-16
[645] Listening Mode Unsupported Feature

### LINKED ISSUE #645: Listening Mode can be toggled on when device is unsupported
**Describe the bug**

The debug menu and in-app settings both permit the user to enable Listening Mode when the device is configured for a non-English language. This should not be possible. 

**To Reproduce**
1. Set the device to a supported non-English language (e.g. German)
2. Open the debug menu
3. Attempt to enable the feature flag  

**Expected behavior**
Toggles for the feature should be OFF and disabled/grayed out so the user can't enable them.
In the settings UI there should be a footer with copy explaining the feature's requirements to indicate why the toggle is disabled.

**Actual behavior**
Currently the toggles behave as though everything is normal

**Device Information**
- Affects all supported devices/OS versions

jsmorgan42:
closes #645

# Description of Work
When a device does not support listening mode (if the device's language is not set to English or if Siri is disabled), the listening mode toggle will now be disabled and have an explanation footer explaining the feature is not supported on the device.

---

### DIFF
--- a/Vocable/AppConfig/ListenModeFeatureConfiguration.swift
+++ b/Vocable/AppConfig/ListenModeFeatureConfiguration.swift
@@ -1,5 +1,5 @@
 //
-//  ListeningModeConfig.swift
+//  ListenModeFeatureConfiguration.swift
 //  Vocable
 //
 //  Created by Chris Stroud on 4/27/22.
@@ -36,7 +36,7 @@ final class ListenModeFeatureConfiguration: ObservableObject {
 
     // The user's preference for whether the overall listening mode feature is enabled
     @PublishedDefault(.listeningModeEnabledPreference)
-    var listeningModeEnabledPreference: Bool = ListenModeFeatureConfiguration.deviceSupportsListeningMode()
+    var listeningModeEnabledPreference: Bool = ListenModeFeatureConfiguration.deviceSupportsListeningMode
 
     // The user's preference for whether the hotword feature is enabled
     @PublishedDefault(.listeningModeHotWordEnabledPreference)
@@ -57,7 +57,7 @@ final class ListenModeFeatureConfiguration: ObservableObject {
                 lhs.2 == rhs.2
             }
             .sink { [weak self] (isFlagEnabled, isListeningModeEnabled, isHotWordEnabled) in
-                let modeEnabled = isFlagEnabled && isListeningModeEnabled && ListenModeFeatureConfiguration.deviceSupportsListeningMode()
+                let modeEnabled = isFlagEnabled && isListeningModeEnabled && ListenModeFeatureConfiguration.deviceSupportsListeningMode
                 let hotwordEnabled = modeEnabled && isHotWordEnabled
                 self?.isEnabled = modeEnabled
                 self?.isHotWordEnabled = hotwordEnabled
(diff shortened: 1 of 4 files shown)


Vocable/AppConfig/ListenModeFeatureConfiguration.swift
```diff
@@ -83,7 +83,7 @@ final class ListenModeFeatureConfiguration: ObservableObject {
         }
     }
 
-    private static func deviceSupportsListeningMode() -> Bool {
+    static var deviceSupportsListeningMode: Bool {
```
moyerr, line 86: Love this 🚀

Vocable/Features/Settings/ListeningMode/ListeningModeViewController.swift
```diff
@@ -142,13 +143,25 @@ final class ListeningModeViewController: VocableCollectionViewController {
...
+                }
+            } else {
+                let model = UIDevice.current.localizedModel
+                let systemName = UIDevice.current.systemName
+                let systemVersion = UIDevice.current.systemVersion
+                let siriName = "Siri"
+
+                let format = String(localized: "settings.listening_mode.device_unsupported_explanation_footer")
+
+                text = String(format: format, model, systemName, systemVersion, siriName)
```
moyerr, line 162: This feels like something that should be part of the dataSource itself. Like maybe `.unsupported` could be its own section?

Also, should we be using `.localizedStringWithFormat(_:...)` here?

jsmorgan42: That's a good point, I'll look into putting this logic into the data source. 

In terms of `localizedStringWithFormat`, Chris had said it isn't necessary to use unless we're dealing with something like money or measurements I believe.

jsmorgan42: Paired on this and decided to keep it the way it is since the section is associated with an item AND and a footer

rhsalas3:
@SashankPatel 

We should discuss this fix with the team. In my testing I have come across some use cases that we should consider.

In particular, on a fresh install we are still presented with the 'Grant Access' flow but cannot tap the button because the device's settings do not support Listen mode. We should consider gating the user at this point as well. I spoke with @jsmorgan42 about this and we agreed it would be best to write up my concerns as separate tickets considering his allocation status and everyone's PTO these next couple of days. 

I will write them up and share them:
#672

rhsalas3:
@jsmorgan42 @SashankPatel @moyerr @Clstroud 

Per a conversation in our standup on 05/18:
- the expected behavior may need to be that the feature is OFF altogether. In other words, we hide the 'Listen' category and 'Listening Mode' settings. 
     - This solves our issue of what happens for a user that installs the app fresh or updates from a version that did not already have listen mode. An issue tracked by ticket #672 
- In the case that a user has already enabled listening mode from a supported language, and then toggles their device to an unsupported language, we would still want to present the "device is not supported" message.

moyerr (approved):

rhsalas3:
Per conversation on 5/26/2022:

We will keep this implementation; merging the PR once it's been re-tested.

For the flow where a user has not yet granted permission to Listen Mode, we will show a new screen that shows both the "not supported" message and the "when listening mode is enabled" description. This will be implemented in #672

rhsalas3:
Re-tested an can confirm our expected behavior, given recent conversations about this feature. @moyerr said he approved this PR and that I am free to merge when tested. Therefore I am going to go ahead and merge. 👍

PR #681, opened 2022-05-26
[677] Listening Mode Responses Update Every New Phrase

### LINKED ISSUE #677: Listen Mode - `Share with Developer` function should be updated after each new phrase
**Acceptance Criteria:**
- For each new phrase, user should get a refreshed `Share with Developer` button. 
- Pressing the "Share with Developer" button should share the most recent phrase spoken.
- We should refresh with a new phrase spoken, even if it would result in the same responses (i.e. subsequent Yes/No questions)

**Steps to Reproduce Actual Behavior**:
1) Go to Listen Mode
2) Say "Test"
3) As expected you will get the sounds complicated error
4) Tap on `Share with Developer`
5) As expected you will get the Thank you checkmark
6) Then, say "Happy"
7) Although you say a new phrase, the `Share with Developer` button is not updated to send the new phrase

jsmorgan42:
closes #677

# Description of Work
This change fixes an issue where the `Share with Developers` button would only send the phrase that last triggered a new response state (i.e. sounds complicated, yes/no). For this change, a new view animates in whenever a new phrase is finalized, even if it's the same state as before.

---

### DIFF
--- a/Vocable/Features/Voice/ListeningResponseViewController.swift
+++ b/Vocable/Features/Voice/ListeningResponseViewController.swift
@@ -103,12 +103,6 @@ final class ListeningResponseViewController: VocableViewController {
     }
 
     private func setContent(_ content: Content, animated: Bool = true) {
-
-        if self.contentViewController != nil {
-            guard content != self.content else {
-                return
-            }
-        }
         let previousContent = self.content
         self.content = content
         let outgoingTransition: TransitionStyle


rhsalas3:
The screen reanimates with each question so that the `Share with developers` button is reset. Triggering the data sends only the most recent phrase/response instead of the first phrase that triggered a response state that is the same through a sequence of similar questions.

The fix looks good to me.

moyerr (approved):
🚀

PR #684, opened 2022-05-26
Enable head gaze interactions in SwiftUI

moyerr:
This PR provides the foundation for enabling head-gaze interactions in SwiftUI view hierarchies. The main entry point for this style of interaction is the `GazeButton`, which is designed to have a similar API to the built-in `Button` type.

This PR also includes DocC-compatible documentation comments on many of the new types and methods, which can be viewed in the the documentation window:

Screen Shot 2022-06-01 at 3 44 11 PM

### Notes for Reviewers

There are a lot of changes in here, many of which are documentation comments, so I wanted to just give a quick outline of the most important components that comprise this feature:

- `GazeButton` - The main component for utilizing head-gaze interactions in SwiftUI
  - `BridgedGazeableButton` - The UIKit component that backs the `GazeButon`
  - Supporting types, such as:
    - `ButtonRole`
    - `ButtonState`
    - `GazeButtonStyle` & `GazeButtonStyleConfiguration`
- `GazeableAlert` -  A bridged version of the app's existing `GazeableAlertViewController` 

**NOTE**: Since this is a foundational PR, no SwiftUI views are actually added to the app yet. However, as an example, I did create a SwiftUI version of the Edit Category Detail screen. I recommend checking out `EditCategoryDetail.swift` to get an idea of what's possible with these foundational pieces. If you'd like to see it in the app (iPad recommended), navigate to `EditCategoriesViewController.swift` and replace line 236 with the following lines:

```Swift
if #available(iOS 15, *) {
  destination = UIHostingController(
    rootView: EditCategoryDetail(objectId: objectID)
      .environment(\.managedObjectContext, NSPersistentContainer.shared.viewContext)
  )
}
```


Vocable/AppDelegate.swift
```diff
@@ -11,6 +11,7 @@ import CoreData
 import Combine
 import AVFoundation
 import VocableListenCore
+import SwiftUI
```
jsmorgan42: Unused import

```suggestion
```

Vocable/Features/Settings/EditCategories/EditCategoriesViewController.swift
```diff
@@ -6,9 +6,10 @@
 //  Copyright © 2020 WillowTree. All rights reserved.
 //
 
-import UIKit
 import Combine
 import CoreData
+import SwiftUI
+import UIKit
```
jsmorgan42: Another unused import

```suggestion
import Combine
import CoreData
import UIKit
```

Vocable/SwiftUI/EditCategoryDetail.swift
```diff
@@ -0,0 +1,152 @@
+//
+//  ExampleVocableView.swift
```
jsmorgan42: Comment seems to still have the old file name

```suggestion
//  EditCategoryDetail.swift
```

Vocable/AppDelegate.swift
moyerr: Ah good catch!

Vocable/SwiftUI/EditCategoryDetail.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    request.predicate = NSPredicate(format: "(SELF = %@)", categoryId)
+    request.sortDescriptors = []
+    request.fetchLimit = 1
+
+    return request
+}
+
+@available(iOS 15, *)
+struct EditCategoryDetail: View {
```
jsmorgan42, line 28: It might be useful to add a comment at the top of this class just saying it's an example view utilizing `GazeButton`

moyerr: Good point, maybe I could also add a deprecation tag or something to indicate that it shouldn't be used in production code

Vocable/SwiftUI/Gaze Button/ButtonState.swift
```diff
@@ -0,0 +1,26 @@
...
+    /// Highlighted state of a button
+    ///
+    /// A button becomes highlighted when a touch or gaze event
+    /// enters the button's bounds, and it loses that highlight
+    /// when there is a touch-up event or when the touch/gaze event
+    /// exits the button's bounds.
+    static let highlighted  = Self(rawValue: 1 << 0)
+
+    /// Selected state of a button
+    static let selected     = Self(rawValue: 1 << 1)
```
jsmorgan42, line 25: Could probably also explain how a button becomes selected like you did for the highlighted state above just to be consistent. i.e. talking about how it becomes selected after the button is in a highlighted state for a specified amount of "hover time."

jsmorgan42: Could probably also explain how a button becomes selected like you did for the highlighted state above just to be consistent. i.e. talking about how it becomes selected after the button is in a highlighted state for a specified amount of "hover time."

Vocable/SwiftUI/Gaze Button/ButtonRole.swift
```diff
@@ -0,0 +1,30 @@
...
+    static let cancel = ButtonRole("cancel")
+
+    /// A role that indicates a destructive button.
+    ///
+    /// Use this role for a button that deletes user data,
+    /// or performs an irreversible operation. A destructive
+    /// button signals by its appearance that the user should
+    /// carefully consider whether to tap the button.
+    static let destructive = ButtonRole("destructive")
+}
```
jsmorgan42, line 30: Why not just use an enum here? Is this to just mirror what Apple is doing for their `ButtonRole` object?

Vocable/SwiftUI/Gaze Button/DefaultGazeButtonStyle.swift
```diff
@@ -0,0 +1,47 @@
...
+                    (state.contains(.highlighted) || !isEnabled) ? 0.5 : 1
+                )
+        }
+    }
+}
+
+extension GazeButtonStyle where Self == DefaultGazeButtonStyle {
+    /// The default button style
+    static var automatic: DefaultGazeButtonStyle { .init() }
+}
```
jsmorgan42, line 47: Could you not just say `extension DefaultGazeButtonStyle` 🤔

Vocable/SwiftUI/Gaze Button/GazeButton.swift
```diff
@@ -0,0 +1,206 @@
...
+///
+///     GazeButton(action: signIn) {
+///         Text("Sign In")
+///     }
+///
+/// For the common case of text-only labels, you can use the convenience
+/// initializer that takes a title string or `LocalizedStringKey` as its first
+/// parameter, instead of a trailing closure:
+///
+///     GazeButton("Sign In", action: signIn)
```
jsmorgan42, line 29: I like it 💯

Vocable/SwiftUI/Gaze Button/GazeButton.swift
```diff
@@ -0,0 +1,206 @@
...
+///
+/// The button's role is used to style the button appropriately. For
+/// example, a destructive button in the default style appears with
+/// a red foreground color:
+///
+/// ![A screenshot of a menu that contains the four items Cut, Copy,
+/// Paste, and Delete. The last item uses a foreground color of red.](GazeButton-Roles)
+///
+/// If you don't specify a role for a button, an appropriate default
+/// appearance is used.
```
jsmorgan42, line 47: By an `an appropriate default appearance`, does this just mean the default styling you would get with a normal SwiftUI button?

Vocable/SwiftUI/Gaze Button/ButtonRole.swift
moyerr: Yep, this was meant to be an exact drop-in replacement for Apple's `ButtonRole` (since theirs is iOS 15+ and our `GazeButton` can work on iOS 14+), which is why it is implemented this way. It is a good question though, I'm not sure why they implemented it as a struct 🤔

Vocable/SwiftUI/Gaze Button/DefaultGazeButtonStyle.swift
moyerr: Great question - the reason this is an extension on `GazeButtonStyle` is so the dot-syntax (`.automatic`) can be used anywhere a `GazeButtonStyle` type is expected, like in the `.gazeButtonStyle(_:)` view modifier. An added benefit is that leveraging this dot-syntax helps discoverability, since you can just start typing with a dot and get some suggestions, rather than needing to remember the actual type names (like `DefaultGazeButtonStyle`).

moyerr: All of SwiftUI's control styles are done in a similar way. You'll even see a suggestion for this pattern in the `ToggleStyle` docs: https://developer.apple.com/documentation/swiftui/togglestyle

Vocable/SwiftUI/Gaze Button/GazeButtonStyleConfiguration.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    /// A type-erased label of a ``GazeButton``.
+    struct Label: View {
+        private var view: AnyView
+        var body: some View { view }
+
+        fileprivate init<V: View>(_ view: V) {
+            self.view = AnyView(view)
+        }
+    }
```
jsmorgan42, line 41: If this underlying view could technically be anything, would it make more sense to name this something like `Content` instead?

Vocable/SwiftUI/UIKit Bridge/BridgedGazeableButton.swift
```diff
@@ -0,0 +1,86 @@
...
+        isHighlighted = false
+    }
+
+    override func gazeCancelled(_ gaze: UIHeadGaze, with event: UIHeadGazeEvent?) {
+        guard isEnabled else { return }
+
+        gazeStartDate = nil
+        isSelected = false
+        isHighlighted = false
+    }
```
jsmorgan42: This doesn't seem to quite match what's happening in `GazeButton` where the start date is set to the `.distantFuture` instead of `nil`. Is there a reason for this change? It also looks like this makes `gazeEnded()` and `gazeCancelled` accomplish the same exact thing.

Vocable/SwiftUI/Gaze Button/ButtonRole.swift
```diff
@@ -0,0 +1,30 @@
+//
+//  ButtonRole.swift
+//  Vocable
+//
+//  Created by Robert Moyer on 4/7/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+/// A value that describes the purpose of a button.
+@available(iOS, obsoleted: 15, message: "Please use the built-in SwiftUI.ButtonRole type.")
```
pg8wood, line 10: 💯

Vocable/SwiftUI/Gaze Button/DefaultGazeButtonStyle.swift
pg8wood: Great question and great answer. TIL :100:

Vocable/SwiftUI/Gaze Button/GazeButton.swift
moyerr: Yes, for the default style. But really, this means for _any_ style, the lack of a role will result in a "normal" looking button, whatever normal means for that specific style.

Vocable/SwiftUI/Gaze Button/GazeButtonStyleConfiguration.swift
moyerr: I think that's a sensible name, but I went with `Label` to match the convention set by the existing SwiftUI controls:

- https://developer.apple.com/documentation/swiftui/buttonstyleconfiguration/label-swift.struct
- https://developer.apple.com/documentation/swiftui/togglestyleconfiguration/label-swift.struct
- https://developer.apple.com/documentation/swiftui/menustyleconfiguration/label

Vocable/SwiftUI/UIKit Bridge/BridgedGazeableButton.swift
moyerr: I think having `gazeEnded()` and `gazeCancelled()` do the same thing is logically correct - in both cases, the gaze event is no longer active, and the state needs to be reset. I could extract the state reset code out to a method and just call that method in each case.

As far as `.distantFuture` vs. `nil`, I don't think there is logical difference since the `gazeMoved()` method will exit early in either case. I guess my thinking here was that, if there is no active gaze event, it seems clearer to me if we `nil` it out rather than set it to an arbitrary value that we deem invalid.

I'd love to have @Clstroud check my assumptions here

Vocable/SwiftUI/Gaze Button/GazeButtonStyleConfiguration.swift
```diff
@@ -0,0 +1,58 @@
...
+/// the ``GazeButtonStyle`` protocol, you implement the
+/// ``GazeButtonStyle/makeBody(_:)`` method. That method takes a
+/// `GazeButtonStyleConfiguration` input that has the information you need
+/// to define the behavior and appearance of a ``GazeButton``.
+///
+/// The configuration structure's ``label-swift.property`` reflects the
+/// button's content, which might be the value that you supply to the
+/// `label` parameter of the ``GazeButton/init(minimumGazeDuration:role:action:label:)`` initializer.
+/// Alternatively, it could be another view that SwiftUI builds from an
+/// initializer that takes a string input, like ``GazeButton/init(_:minimumGazeDuration:role:action:)``.
```
pg8wood: Looks like this function signature changed and this is now generating a warning

```suggestion
/// initializer that takes a string input, like ``GazeButton/init(minimumGazeDuration:role:action:label:)``.
```

Vocable/SwiftUI/Gaze Button/VocableGazeButtonStyle.swift
```diff
@@ -0,0 +1,81 @@
...
+                    .cornerRadius(8)
+                    .overlay {
+                        if shouldHighlight {
+                            RoundedRectangle(cornerRadius: 8)
+                                .stroke(Color(UIColor.cellBorderHighlightColor), lineWidth: 4)
+                        }
+                    }
+                    .scaleEffect(shouldHighlight ? 0.95 : 1)
+            } else {
+                // Fallback on earlier versions
```
pg8wood: 🤯 what the heck is the result builder coming up with on < iOS 15? Looks like we'll want to do something here, since now it doesn't seem to do much

Vocable/SwiftUI/GazeableAlert.swift
```diff
@@ -0,0 +1,89 @@
...
+                GazeableAlertRepresentable(
+                    message: message,
+                    actions: actions
+                )
+                .edgesIgnoringSafeArea([])
+                .transition(.scale)
+            }
+            .ignoresSafeArea()
+        }
+    }
```
pg8wood, line 54: TIL that this works

Vocable/SwiftUI/UIKit Bridge/BridgedGazeableButton.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+//  Created by Robert Moyer on 4/1/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import Combine
+import UIKit
+
+class BridgedGazeableButton: UIButton {
+    var minimumGazeDuration: TimeInterval = 2
```
pg8wood: Does this override the user's gaze settings?

Vocable/SwiftUI/UIKit Bridge/BridgedGazeableButton.swift
```diff
@@ -0,0 +1,88 @@
...
+    // MARK: Gaze Overrides
+
+    override func gazeBegan(_ gaze: UIHeadGaze, with event: UIHeadGazeEvent?) {
+        guard isEnabled else { return }
+
+        gazeStartDate = Date()
+        isHighlighted = true
+    }
+
+    override func gazeMoved(_ gaze: UIHeadGaze, with event: UIHeadGazeEvent?) {
```
pg8wood, line 59: Any reason to not call the `super` implementations for this type's gaze events?

pg8wood (commented):
Awesome job mirroring the SwiftUI API. If Apple ever Sherlocks this functionality, you've made it very easy for them 😉

I was able to play around with the `EditCategoryDetailView` and make my own gaze-able buttons. Looks fantastic! 


It seems that the `BridgedGazeableButton` breaks VoiceOver somehow--it only announces everything as "button".

Vocable/SwiftUI/Gaze Button/GazeButtonStyleConfiguration.swift
moyerr: Nice catch! I think the reason it's throwing a warning is because this is an ambiguous reference since the initializers that accept a `LocalizedStringKey` and a `StringProtocol` each have the same signature. DocC documentation tells me that in this case it should generate a random identifier to disambiguate, but for some reason it's not doing that here 🤔 - I will look into this 🕵️‍♂️

Vocable/SwiftUI/GazeableAlert.swift
moyerr: Yes, as I understand it, the `ViewBuilder` will return an `EmptyView` for code paths where nothing is explicitly returned.

Vocable/SwiftUI/Gaze Button/VocableGazeButtonStyle.swift
moyerr: Oh, nice catch! I meant to get back to this, but never did 🤦‍♂️ 

A small refactor and the iOS 15 constraint is no longer needed 👍

Vocable/SwiftUI/UIKit Bridge/BridgedGazeableButton.swift
moyerr: Just an oversight, thanks for catching this! 💯

Vocable/SwiftUI/UIKit Bridge/BridgedGazeableButton.swift
Clstroud: 1) Yes - consider the difference between touch (ephemeral) and gaze (persistent). We're mirroring the touch event API, so we're sort of forcing one thing to behave like another.
2) It's been a while, but I _think_ the reason for `.distantFuture` was to have a date way off in the future that would not just be some magic number, yet mathematically give the same results without needing to unwrap an optional all the time. `nil` is probably clearer, but my recollection is that the branching for all the guards was getting annoying. Not offended by it being optional.

Vocable/SwiftUI/UIKit Bridge/BridgedGazeableButton.swift
moyerr: We use the one supplied by the `GazeButton`, which defaults to the user's setting. Probably good to have this default to the same value. Although, I think best-case scenario would be to have these values update when the `AppConfig`'s value updates. I'll try to think of a nice way to solve for that, while maintaining a flexible API for supplying a custom value.

Vocable/SwiftUI/Gaze Button/GazeButtonStyleConfiguration.swift
moyerr: Fixed ✅

jsmorgan42 (approved):
LGTM ship it 😎 🚢 🌊

PR #685, opened 2022-05-26
[672] Unsupported Listening Mode Before Microphone Access

### LINKED ISSUE #672: Listening Mode: Prevent "Grant Access" Flow for Non-English Languages
### Description
We have implemented a message to the user that their device does not support listening mode, when that device is set to a non-English language (issue #645 -> PR #663 ). On a fresh install of the app the user must complete the flow to grant access to Speech Recognition and the Microphone before Listening Mode is enabled. In this scenario, of a device that is set to a non-English language, they are presented with an unresponsive and unnecessary grant-access-to-Speech-Recognition screen. We should consider presenting our "device is not supported" message here as well, instead of the grant access flow. 

Furthermore, it may be a good idea to include our "when listening mode is enabled" description along with the message. This way the user can still be informed about what the feature is.

Update: Question for the team...what is the expected behavior when your device is set to an unsupported language, which defaults to English?

**Acceptance Criteria**:
- [x] While the device is set to a non-English language, the user is presented with the "device is not supported" message instead of the grant access flow on the Settings > Listening Mode screen, and Settings > Categories and Phrases > 'Listen' category details screen.
    - [x] on a fresh install of Vocable
    - [x] when upgrading from a version that does not have listening mode to one that does

- [x] ~~The "when listening mode is enabled" description text accompanies the "devices is not supported" message where the latter is displayed instead of the "grant access" flow~~. We decided not to include the description, opting for MVP whereby we only display the "device is not supported" text.

- [x] The "device is not supported" message continues to display as expected, when the user has already granted access to listening mode but then toggles their device to a non-English language; per our implementation of #645 

**Design**:

We will probably need designs or at least approved copy.

jsmorgan42:
closes #672

# Description of Work
A new view now appears in the listening mode settings screen for a user that has not yet granted microphone access but how their device set to an unsupported language (any language other than English).

## Notes to Test
- Due to the nature of localization, when in a non-english language, you will only see the localized variable name rather the actual text like the screenshot below.
- We should also verify that the listen mode does not show up anywhere else in the app when in the non-english mode other than the settings screen

---

### DIFF
--- a/Vocable/Features/Voice/ListeningResponseEmptyStateViewController.swift
+++ b/Vocable/Features/Voice/ListeningResponseEmptyStateViewController.swift
@@ -16,6 +16,7 @@ extension EmptyStateView {
 
 enum ListeningEmptyState: EmptyStateRepresentable, Equatable {
 
+    case listeningModeUnsupported
     case listeningResponse
     case speechServiceUnavailable
     case speechPermissionDenied
@@ -26,6 +27,8 @@ enum ListeningEmptyState: EmptyStateRepresentable, Equatable {
 
     var title: String {
         switch self {
+        case .listeningModeUnsupported:
+            return String(localized: "listening_mode.empty_state.unsupported.title")
         case .microphonePermissionDenied:
             return String(localized: "listening_mode.empty_state.microphone_permission_denied.title")
         case .microphonePermissionUndetermined:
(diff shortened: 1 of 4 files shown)


Vocable/Features/Voice/ListeningResponseEmptyStateViewController.swift
```diff
@@ -80,6 +92,15 @@ enum ListeningEmptyState: EmptyStateRepresentable, Equatable {
...
     }
+
+    var yOffset: CGFloat? {
+        switch self {
+        case .listeningModeUnsupported:
+            return -64
+        default:
+            return nil
+        }
+    }
```
moyerr, line 103: Interesting - did the standard center alignment not look right for this empty state?

moyerr (approved):
Looks good to me, just a small question about the `yOffset` business

rhsalas3:
Tested and verified on the following devices:
- iPhone 8 Plus, iOS 14.8.1
- iPhone X, iOS 14.2
- iPad Pro, iOS 15.3.1

Vocable/Features/Voice/ListeningResponseEmptyStateViewController.swift
jsmorgan42: Yeah, the empty state without a button in this settings screen definitely looked off without some vertical offset

PR #693, opened 2022-06-01
#688 | Speech Services Unavailable UI Updates

### LINKED ISSUE #688: Include Dictation and Siri requirements for Listening Mode permissions/setup
### Description:
Vocable leverages a device's Dictation and Siri capabilities for Listening Mode. The long term plan is to include information about this in the onboarding flow for how to setup listening mode properly; which device settings/permissions are required, etc. Until then we should update the copy for listening mode's "speech services unavailable" error message to inform users of what is needed for listening mode to function properly, instead of "please try again later".

### Proposed New Message
"Speech services require the following features to be enabled on your device: Siri and Dictation. Please check your device settings as one or all of these may be disabled."

Repro Steps:
_Device settings differ based on OS version and hardware, these steps work on an iPad Pro running 15.3.1_
1. Open the `Settings` app
2. Open the `Screen Time` settings
3. Open Content & Privacy Restrictions > Allowed Apps > `Siri & Dictation`
4. Toggle the feature OFF
5. Launch or re-open Vocable
6. On the Main Screen, navigate to the 'Listen' category
7. Note the error message (screenshot below)

### Acceptance Criteria:
- [this AC still needs crafting] Mention that for listening mode to work best, Settings > General > Keyboard > Enable Dictation should be on
- [this AC still needs crafting] Mention that for listening mode to work best, Settings > Siri & Search > "Hey Siri" and Press top button for Siri are enabled
    - Note: having at least one of this means Siri is enable on the device, but both of them off means Siri is not enabled.

**Design**:

Referencing the proposed onboarding flow, which is still a work-in-progress: https://www.figma.com/file/mNrwUygVhTmuWnKNHHkenR/Vocable---Current?node-id=2199%3A0

**Screenshot of error message as of 1.4**:
SpeechServicesUnavailable_ListeningMode_error

jsmorgan42:
closes #688

# Description of Work
When speech services are unavailable, the error state now tells the user to enable Dictation & Siri ~~and also now contains a button link the the device settings.~~ **The settings button has been removed since it might be confusing for the user to be deep linked to the detail page of the Vocable app**

## Notes to Test
~~- When linking the user to the settings, Apple will automatically try to deep link to the Vocable app settings detail page. There is no way to deep link to the main device settings screen.~~ **No longer relevant since settings button was removed.**

---


Vocable/Extensions/URL+Helpers.swift
```diff
@@ -0,0 +1,20 @@
...
+extension URL {
+
+    static func openSettingsURL() {
+        if let url = URL(string: UIApplication.openSettingsURLString) {
+            if UIApplication.shared.canOpenURL(url) {
+                UIApplication.shared.open(url, options: [:], completionHandler: nil)
+            }
+        }
+    }
+}
```
Clstroud: This feels like a strange behavior for `URL` to statically provide 🤔 I could see it returning the URL itself, but it feels more appropriate for this to be an extension on something like `UIApplication` itself, perhaps?

Vocable/Features/Voice/AudioPermissionPromptController.swift
```diff
@@ -71,11 +67,7 @@ final class AudioPermissionPromptController {
             return nil
         case .denied: // Need to go to settings
             return .init(state: .microphonePermissionDenied, action: {
```
Clstroud, line 69: I think these actions could be [curried](https://thoughtbot.com/blog/introduction-to-function-currying-in-swift) if you wanted to

Clstroud (approved):
Couple of suggestions, but otherwise 💰 💯

Vocable/Extensions/URL+Helpers.swift
jsmorgan42: Yeah I also felt a little weird about it. Wasn't sure exactly where to put it. I think an extension on `UIApplication` definitely makes more sense, I'll change that

Vocable/Features/Voice/AudioPermissionPromptController.swift
jsmorgan42: Interesting, I'll look into this later today if I have some extra time

CananArikan:
Thanks for the updates! Test verified!

Device: iPad Pro
OS: iOS 15.5

PR #701, opened 2022-06-07
Bugfix for Listening Mode Empty View State

jsmorgan42:
# Description of Work
There was an issue where the listening mode screen would sometimes say it's listening when it should say "Speech Services Unavailable" This was because when there was a nil permissions state, the default was set to the listening response empty state instead of the speech services unavailable state.

---

### DIFF
--- a/Vocable/Features/Voice/AudioPermissionPromptController.swift
+++ b/Vocable/Features/Voice/AudioPermissionPromptController.swift
@@ -31,7 +31,6 @@ final class AudioPermissionPromptController {
         let micStatusPublisher = controller.$microphonePermissionStatus.dropFirst().removeDuplicates()
         let speechStatusPublisher = controller.$speechPermissionStatus.dropFirst().removeDuplicates()
         Publishers.CombineLatest(micStatusPublisher, speechStatusPublisher)
-            .dropFirst()
             .receive(on: DispatchQueue.main)
             .sink { [weak self] (micStatus, speechStatus) in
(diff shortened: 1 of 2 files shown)


Vocable/Features/Voice/AudioPermissionPromptController.swift
```diff
@@ -31,6 +31,7 @@ final class AudioPermissionPromptController {
         let micStatusPublisher = controller.$microphonePermissionStatus.dropFirst().removeDuplicates()
         let speechStatusPublisher = controller.$speechPermissionStatus.dropFirst().removeDuplicates()
         Publishers.CombineLatest(micStatusPublisher, speechStatusPublisher)
+            .dropFirst()
```
Clstroud: Just wanted to double check that this is intentional, since it's effectively dropping twice? Once for each publisher, then once again for the aggregate?

rowildcard:
@jsmorgan42  Sorry to say this broke the Grant Access button. Stop by when you have a chance.

Vocable/Features/Voice/AudioPermissionPromptController.swift
jsmorgan42: Yeah this was just because it was making `authorizationStatusDidChange()` be called twice. Once in initialization & once in this publisher when first entering the listening mode screen. However, I'm now realizing that it might causes problems if the state is changed when they're still in the same listening mode screen. I can remove this change.

Clstroud (approved):
Makes sense 👍🏻 💯

PR #703, opened 2022-06-13
Fixing Mixpanel Prod Token Default

jsmorgan42:
# Description of Work
The mixpanel token was defaulting to the `staging` token even when it wasn't a debug build, causing mixpanel data to never be sent to the prod environment. Also increasing version number since new changes would be added to the production build.

## Notes to Test
This needs to be merged before it can be merged since it requires a TestFlight build to be created to check that it's sending up data to Mixpanel prod.

---


SashankPatel (approved):
Looks Good.

PR #708, opened 2022-06-21
Added Custom Category App Restart Tests

### LINKED ISSUE #707: [Automation] [Custom] Verify customization persists after restart
Add an automation test to verify that customization persists after restart.
[TC-47: [Custom] Verify customization persists after restart](https://vocable.wtadev.com/case/47/)

**Pre-requisites**:
The following customizations have been performed:

- 2 custom categories have been added.
- Custom phrases have been added to both categories.
- One of the custom categories is hidden.
- One of the custom categories is reordered.

**The user restarts the app**:

1. Force quit the app.
2. Reopen the app.

**Acceptance Criteria**:
All user customization persists after the app restarts.

CananArikan:
closes #707 

**Description:**
Added automation tests to verify that Custom Category customizations persist after restarting the app.

### DIFF
--- a/VocableUITests/Tests/CustomCategoryAppRestartTests.swift
+++ b/VocableUITests/Tests/CustomCategoryAppRestartTests.swift
@@ -53,7 +53,6 @@ class CustomCategoryAppRestartTests: XCTestCase {
         XCTAssertTrue(CustomCategoriesScreen.phraseDoesExist(phrase))
     }
     
-    // 16 seconds
     func testHideCategoryPersists() {
         // Verify that custom category exists
         XCTAssertTrue(SettingsScreen.doesCategoryExist(firstCustomCategory))
@@ -107,7 +106,7 @@ class CustomCategoryAppRestartTests: XCTestCase {
         
         // Reorder custom categories, move first custom category to the end of the list
         SettingsScreen.locateCategoryCell(firstCustomCategory)
-        firstCategory.buttons[.settings.editCategories.moveDownButton].tap()
+        firstCategory.buttons[.settings.editCategories.moveDownButton].tap(afterWaitingForExistenceWithTimeout: 1)
         
         // Verify that first custom category's (last in the list), up button is enabled and down button is disabled
         XCTAssertTrue(firstCategory.buttons[.settings.editCategories.moveUpButton].isEnabled)
--- a/VocableUITests/Tests/CustomCategoryAppRestartTests.swift
+++ b/VocableUITests/Tests/CustomCategoryAppRestartTests.swift
@@ -1,0 +1,137 @@
+//
+//  CustomCategoryAppRestartTests.swift
+//  VocableUITests
+//
+//  Created by Canan Arikan on 6/16/22.
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import XCTest
+
+class CustomCategoryAppRestartTests: XCTestCase {
+    
+    private(set) var firstCustomCategory: String = "First"
+    private(set) var secondCustomCategory: String = "Second"
+    private(set) var phrase: String = "Test"
+
+    override func setUp() {
+        let app = XCUIApplication()
+        app.configure {
+            Arguments(.resetAppDataOnLaunch, .disableAnimations)
+        }
+        continueAfterFailure = false
+        app.launch()
+        
+        // Create a custom category
+        SettingsScreen.navigateToSettingsCategoryScreen()
+        CustomCategoriesScreen.createCustomCategory(categoryName: firstCustomCategory)
+    }
+    
+    func testCategoryAndPhraseCustomizationPersists() {
+        // Verify that custom category exists
+        XCTAssertTrue(SettingsScreen.doesCategoryExist(firstCustomCategory))
+        
+        // Add a custom phrase to the custom category and verify that it exists
+        SettingsScreen.openCategorySettings(category: firstCustomCategory)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
+        CustomCategoriesScreen.addPhrase(phrase)
+        XCTAssertTrue(CustomCategoriesScreen.phraseDoesExist(phrase))
+     
+        // Restart the app
+        app.terminate()
+        app.activate()
+        
+        // Navigate to Settings Category Screen
+        SettingsScreen.navigateToSettingsCategoryScreen()
+        
+        // Verify that the custom category and phrase persist after restarting
+        SettingsScreen.locateCategoryCell(firstCustomCategory)
+        XCTAssertTrue(SettingsScreen.doesCategoryExist(firstCustomCategory))
+        
+        SettingsScreen.openCategorySettings(category: firstCustomCategory)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
+        XCTAssertTrue(CustomCategoriesScreen.phraseDoesExist(phrase))
+    }
+    
+    // 16 seconds
+    func testHideCategoryPersists() {
+        // Verify that custom category exists
+        XCTAssertTrue(SettingsScreen.doesCategoryExist(firstCustomCategory))
+        
+        // Verify that when the custom category is shown (last in the list), up button is enabled and down button is disabled
+        let shownCategory = SettingsScreen.locateCategoryCell(firstCustomCategory)
+        XCTAssertTrue(shownCategory.buttons[.settings.editCategories.moveUpButton].isEnabled)
+        XCTAssertFalse(shownCategory.buttons[.settings.editCategories.moveDownButton].isEnabled)
+        XCTAssertTrue(shownCategory.element.isEnabled)
+        
+        // Hide the custom category
+        SettingsScreen.openCategorySettings(category: firstCustomCategory)
+        SettingsScreen.showCategoryButton.tap()
+        SettingsScreen.navBarBackButton.tap()
+        
+        // Verify that when the category is hidden, up and down buttons are disabled
+        let hiddenCategory = SettingsScreen.locateCategoryCell(firstCustomCategory)
+        XCTAssertFalse(hiddenCategory.buttons[.settings.editCategories.moveUpButton].isEnabled)
+        XCTAssertFalse(hiddenCategory.buttons[.settings.editCategories.moveDownButton].isEnabled)
+        XCTAssertTrue(hiddenCategory.element.isEnabled)
+        
+        // Restart the app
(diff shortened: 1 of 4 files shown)


VocableUITests/Tests/CustomCategoryAppRestartTests.swift
```diff
@@ -0,0 +1,137 @@
...
+        // Verify that the custom category and phrase persist after restarting
+        SettingsScreen.locateCategoryCell(firstCustomCategory)
+        XCTAssertTrue(SettingsScreen.doesCategoryExist(firstCustomCategory))
+        
+        SettingsScreen.openCategorySettings(category: firstCustomCategory)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
+        XCTAssertTrue(CustomCategoriesScreen.phraseDoesExist(phrase))
+    }
+    
+    // 16 seconds
```
rowildcard: Just a little thing... what does this comment mean?

rowildcard (approved):
Very impressive. I even learned a couple of things! Great work.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
CananArikan: This is the test execution time, sorry I forgot to delete it.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
```diff
@@ -0,0 +1,138 @@
...
+        // Restart the app
+        app.terminate()
+        app.activate()
+        
+        // Navigate to Settings Category Screen
+        SettingsScreen.navigateToSettingsCategoryScreen()
+        
+        // Verify that the custom category and phrase persist after restarting
+        SettingsScreen.locateCategoryCell(firstCustomCategory)
+        XCTAssertTrue(SettingsScreen.doesCategoryExist(firstCustomCategory))
```
rhsalas3, line 49: These two functions called here do the same loop through the pages of categories to locate the category. The difference only being that `locateCategoryCell` returns a query locating the cell, but you're not using the return value. So you should be able to delete line 48 altogether. 

Then you also won't need to add the `@discardableResult` annotation in SettingsScreen...unless you want to keep that there.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
```diff
@@ -0,0 +1,138 @@
...
+//  Copyright © 2022 WillowTree. All rights reserved.
+//
+
+import XCTest
+
+class CustomCategoryAppRestartTests: XCTestCase {
+    
+    private(set) var firstCustomCategory: String = "First"
+    private(set) var secondCustomCategory: String = "Second"
+    private(set) var phrase: String = "Test"
```
rhsalas3, line 15: I don't think it's hurting anything to have these be `private(set) var` but it isn't necessary. Since they aren't being manipulated anywhere in the code and are only accessed within this Class you should just set them with `let` instead of `var`. 

```
let firstCustomCategory = "First"
let secondCustomCategory = "Second"
// etc...
```

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
```diff
@@ -0,0 +1,138 @@
...
+        
+        // Add a custom phrase to the custom category and verify that it exists
+        SettingsScreen.openCategorySettings(category: firstCustomCategory)
+        CustomCategoriesScreen.editCategoryPhrasesButton.tap()
+        CustomCategoriesScreen.addPhrase(phrase)
+        XCTAssertTrue(CustomCategoriesScreen.phraseDoesExist(phrase))
+     
+        // Restart the app
+        app.terminate()
+        app.activate()
```
rhsalas3, line 42: Today I learned you can terminate and re-activate the app without losing control during testing...awesome! 💯  💯 

The only feedback I would give here: since you make these same two calls in each test it makes sense to consolidate the code and make this a helper function. I'd consider putting it here in the test class. 
Something like...
```
private func restartApp() {
    // The two calls needed to terminate and activate the app
}
```

I have seen codebases that utilize a `Utilities` class with these types of functions in them but that's out of scope for this effort since it's just these 3 tests that uses the functionality.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
```diff
@@ -0,0 +1,138 @@
...
+        XCTAssertTrue(hiddenCategory.element.isEnabled)        
+    }
+    
+    func testReorderPersists() {
+        // Verify that first custom category exists
+        XCTAssertTrue(SettingsScreen.doesCategoryExist(firstCustomCategory))
+        
+        // Create second custom category and verify that it exists
+        CustomCategoriesScreen.createCustomCategory(categoryName: secondCustomCategory)
+        sleep(1)
```
rhsalas3, line 95: We really should avoid explicit `sleep` calls in our code. It's better to wait with a `waitForExistence()` call on a known element on the destination screen. For example, you may be able to wait for the back arrow to reappear after the keyboard screen is dismissed:
```
SettingsScreen.navBarBackButton.waitForExistence(timeout: 0.5)
```

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
```diff
@@ -0,0 +1,138 @@
...
+        
+        // Verify that second custom category's (last in the list), up button is enabled and down button is disabled
+        let secondCategory = SettingsScreen.locateCategoryCell(secondCustomCategory)
+        XCTAssertTrue(secondCategory.buttons[.settings.editCategories.moveUpButton].isEnabled)
+        XCTAssertFalse(secondCategory.buttons[.settings.editCategories.moveDownButton].isEnabled)
+        
+        // Reorder custom categories, move first custom category to the end of the list
+        SettingsScreen.locateCategoryCell(firstCustomCategory)
+        firstCategory.buttons[.settings.editCategories.moveDownButton].tap()
+        sleep(2)
```
rhsalas3, line 111: Again here, we want to avoid explicit `sleep` calls in our code. In this case you may be able to "buy yourself time" by calling `.tap(afterWaitingForExistenceWithTimeout: 2)` on the reorder button. The actual wait in the `waitForExistence` function occurs before querying for the element's existence. So you'll get the wait time out of the call regardless of whether or not the reorder button actually disappears/reappears.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
rhsalas3: This is a change I would recommend making.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
```diff
@@ -0,0 +1,138 @@
...
+        
+        // Hide the custom category
+        SettingsScreen.openCategorySettings(category: firstCustomCategory)
+        SettingsScreen.showCategoryButton.tap()
+        SettingsScreen.navBarBackButton.tap()
+        
+        // Verify that when the category is hidden, up and down buttons are disabled
+        let hiddenCategory = SettingsScreen.locateCategoryCell(firstCustomCategory)
+        XCTAssertFalse(hiddenCategory.buttons[.settings.editCategories.moveUpButton].isEnabled)
+        XCTAssertFalse(hiddenCategory.buttons[.settings.editCategories.moveDownButton].isEnabled)
```
rhsalas3, line 74: This code block is common throughout each of these tests, whereby you locate the cell and then make the same assertion on each reorder button. Based on feedback from @Clstroud in the past, on the pagination controls, this is a good opportunity to expand on our [VocableTestAsserts](https://github.com/willowtreeapps/vocable-ios/blob/develop/VocableUITests/VocableTestAsserts.swift) class. Maybe adding something like `VTAssertReorderButtonsAreEnabled` and `VTAssertReorderButtonsAreDisabled` assertion functions. You would not re-use the PaginationArrows struct, instead maybe creating one that is similar for the reorder buttons: `ReorderCategoryArrows`. 

I'm making the assumption that your function(s) would take in the category cell query. Then each time you would locate the cell and make your assertions in these tests, you would make one call:
```
VTAssertReorderButtonsAreEnabled(SettingsScreen.locateCategoryCell(firstCustomCategory))
// OR
VTAssertReorderButtonsAreDisabled(SettingsScreen.locateCategoryCell(firstCustomCategory))
```

Just food for thought.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
rhsalas3: This change I would recommend making.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
rhsalas3: This change I would recommend making.

rhsalas3 (commented):
Good work @CananArikan! 

I did leave some feedback. There are a few things I would recommend changing; I've noted so in the respective comments. Everything else is just optional feedback. 🤪

rhsalas3:
Haha oops, looks like I wasn't fast enough with my feedback before you merged the code. Oh well!

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
CananArikan: Thanks, done!

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
CananArikan: Done.

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
CananArikan: I was planning to have a `Utilities` class for further implementations, it is created!

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
CananArikan: Updated #711!

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
CananArikan: Updated #711!

VocableUITests/Tests/CustomCategoryAppRestartTests.swift
CananArikan: Updated #711!

PR #725, opened 2023-08-23
Gets the app building on CI again

andrewroycarter:
# Description

For some upcoming work I'd like to get builds succeeding again.

There have been some moves internally to prefer using GitHub actions for projects like this and internal projects - for the time being I think remaining on circle is easiest. Once we get some of the work we have planned done and released we can look into a potential change.

# Changes

* Updates circle ci to use current version of Xcode
* Switching to use M1 agents on circle.
* Removes crashlytics build phase
* Resolves some failing UI tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,8 @@ version: 2.1
 jobs:
   build-and-test:
     macos:
-      xcode: 13.3.0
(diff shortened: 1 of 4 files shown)


stevefosterwta (approved):
👍

Clstroud (approved):
🥇

PR #736, opened 2024-05-07
Prospective UI test typing improvements

Clstroud:
* Converts plaintext keyboard keys to use fixed identifiers
* Fixes build warnings from unused return values
* Marginally improves typing speed in tests by referencing key identifiers and relying on static keyboard layout


stevefosterwta (approved):
👍

PR #737, opened 2024-05-07
Add feature flag for smart assist

Clstroud:
Disables the feature for the time being until it can be fully tested before release. Don't want to hold up the voice picker work any longer than necessary.


stevefosterwta (approved):
👍

PR #740, opened 2024-05-08
Migrate Deployment Target to iOS 15.0

### LINKED ISSUE #739: Migrate Deployment Target to iOS 15.0
* Project currently requires iOS 13.0+
* In order to adopt String Catalogs successfully, we need to migrate to iOS 15.0+ so we can drop the `String(localized:)` shim we introduced to (as I recall) avoid the "exporting an empty comment" issue from `NSLocalizedString`. The shim we introduced is breaking Xcode's ability to extract strings automatically because it's not the Foundation implementation of the function signature.

String Catalogs are an important change because they will help us audit our strings much more effectively and ensure we aren't carrying around stale translations or untranslated keys.

Clstroud:
closes #739 

* Bumps the project to iOS 15 and removes relevant obsolete code. 
* Will require a separate effort to migrate to `UIButton.Configuration` to clean up some warnings, but I can throw something in the backlog for that.
* Cleans up a couple of low-hanging warnings


stevefosterwta (approved):
👍

PR #742, opened 2024-05-09
Initial String Catalog Migration

### LINKED ISSUE #741: Migrate localized strings to String Catalogs
Once the iOS 15.0 migration has completed, we can take advantage of String Catalogs in Xcode.

As stated in #739, String Catalogs are an important change because they will help us audit our strings much more effectively and ensure we aren't carrying around stale translations or untranslated keys. I am hopeful it will help the Crowdin import/export process, but even if that doesn't change we should see substantial workflow benefits.

Clstroud:
closes #741 

* Migrate the previous `.strings` files to Xcode String Catalogs
* Prune stale translations that are no longer used
* Remove a couple of translations that should not be translated (a debug setting and an unused Voice settings key)

More work is needed to determine how we want to manage these strings overall. If we shift to a pattern defining them 100% in code, we can leverage the `defaultValue` argument of `String(localized: ...)` and have untranslated strings fall back to the development locale instead of a raw key.


stevefosterwta (approved):
👍

Clstroud:
Merging this so we can see how Crowdin reacts to the changes (if at all)

PR #757, opened 2024-05-20
[756] Add restart copy for Personal Voice empty state

### LINKED ISSUE #756: Add instruction to restart device as Personal Voice workaround
When Personal Voice access is first granted, a system bug appears to prevent personal voices from being returned from `AVSpeechSynthesisVoice.speechVoices()`. Additionally, `AVSpeechSynthesizer.availableVoicesDidChangeNotification` does not seem to fire after granting permission, despite being [indicated explicitly by the docs](https://developer.apple.com/documentation/avfaudio/avspeechsynthesizer/4134505-availablevoicesdidchangenotifica).

Killing Vocable and restarting after granting permissions resolves the issue, and we've verified in the debugger that the system simply will not return personal voices otherwise. Since this behavior is likely to be resolved by a future iOS update, the cleanest course of action is to inform the user that they may need to restart their device if a Personal Voice they expect to see in Vocable is not present.

Clstroud:
closes #756 

Updates the copy to inform the user that they should restart their device if a personal voice is not appearing in Vocable as expected.

### DIFF
--- a/CrowdinExport/en.xliff
+++ b/CrowdinExport/en.xliff
@@ -299,8 +299,12 @@ You can enable access in %1$@ Settings.</target>
         <note>Title of the Personal Voices screen empty state when a user must grant access to the app.</note>
       </trans-unit>
       <trans-unit id="personal_voices.empty_state.no_content.description" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


stevefosterwta (approved):
👍

PR #759, opened 2024-05-20
[752] Fix localization format token with incorrect index

### LINKED ISSUE #752: Incorrect localization token format variable
The `head_tracking_unsupported_footer` localization shows "null" for one of the format variables due to the last value being numbered incorrectly. 

It is currently 
```
This %1$@ on %2$@ %3$@ does not support head tracking.

Head tracking is supported on all devices with a %4$@ camera, and on most devices with %6$@.
```
and should be
```
This %1$@ on %2$@ %3$@ does not support head tracking.

Head tracking is supported on all devices with a %4$@ camera, and on most devices with %5$@.
```

Clstroud:
closes #752

Fix the localization format token that had an incorrect index following recent changes.


stevefosterwta (approved):
👍

PR #775, opened 2024-05-22
[773] Keyboard performance improvements

### LINKED ISSUE #773: Improve keyboard performance
Currently, some minor UX issues with the keyboard can be addressed with a few tweaks:

1. Ensure animations allow user interaction
2. Move away from reloading the collection view each time and instead lean more on cell configuration

These changes should also address the whole-cell fading we see for suggestions updates, which will be an aesthetic improvement.

Clstroud:
closes #773 

* Makes the keyboard noticeably (dare I say dramatically?) more responsive
* You can now do things like double-press a key with fewer errors
* Animations for suggested snippets are now smoother

https://github.com/willowtreeapps/vocable-ios/assets/143916/2c070732-bc2f-4fd2-8899-6191f68bf281


stevefosterwta (approved):
👍 Looks wonderful

PR #777, opened 2024-05-23
[732] - Migrate to GitHub Actions

### LINKED ISSUE #732: Investigate migration to Github Actions
Attempt to migrate the current CircleCI pipeline to Github Actions.

Our localization workflow is no longer working due to branch protections on `develop`, which we don't want to remove. Github Actions should theoretically allow us to enforce localization exports such that they occur on PR branches exclusively, while also enforcing that step as a merge requirement.

peterkos:
closes #732 

## Background

Migrate from CircleCI (`circleci.yml`) to GitHub Actions. Partially to support writing the `.xliff` localization file on protected branches, but partially because it gives us some nice affordances :)

## Contents

The "main" jobs are event-driven:

|          Job          |                    Desc                   |
|-----------------------|-------------------------------------------|
| `on_pull_request.yml` | Run on PRs to develop                     |
| `on_release_push.yml` | Run on new pushes to `release.*` branches |
| `on_crowdin_push.yml` | Run on new pushes to `crowdin.*` branches |

These jobs call composite actions[0] (reusable CI steps) to cleanup callsites, but **also** to make it clear what runs when. "On pull requests, XYZ steps are run" kinda thing.

|     Action     |                   Desc                   |
|----------------|------------------------------------------|
| `run_lane`     | Runs the input fastlane `lane`           |
| `xliff_import` | Runs the xliff_import lane after pulling |
| `xliff_export` | Commit the translation file if changed   |

## Notes

<details>
<summary>[0] What's a composite action?</summary>

[This StackOverflow post](https://stackoverflow.com/a/71570847/1431900) goes into good detail, but as a brief summary, there's three kinds of "reusable thing" in GHActions:
- Dispatch workflow
- Reusable workflow
- Composite action

Dispatch is a little hacky (can only run once in default branch, only thru API). Reusable workflows are ephemeral (don't maintain state per-job), so we'd need to upload/download artifacts between each step. Composite actions run in the same job! :) But they have two fun quirks:
- Its yaml must be titled with `{name}/aciton.yml`
- Any `run: ...` actions must specify a `shell: ...`, even though that is officially an optional property ☔ 

</details>


peterkos:
Remaining:
- [ ] Resolve the corrupted SPM resolution file error (!)
- [ ] Validate artifacts work as expected
- [ ] Validate fastlane test annotations (fancy GitHub UI) are showing as expected in a PR
- [ ] Validate Crowdin translation flow work as expected
- [ ] Add `MATCH_PASSWORD` key to secrets if not already there (missing perms 🥺)

.github/workflows/actions/run_lane/action.yml
```diff
@@ -0,0 +1,20 @@
+name: 'Run Lane'
+description: 'Run a fastane lane and upload artifacts'
```
moyerr: Not sure if this is personal preference, but just wanted to point out that string delimiters are not strictly necessary in YAML 😄

This [StackOverflow answer](https://stackoverflow.com/a/22235064) has some helpful info about when you may _need_ delimiters (and what kind). If you're unsure about actions syntax, I highly recommend using the [GitHub actions VSCode extension](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions) if you don't already have it!

.github/workflows/actions/run_lane/action.yml
```diff
@@ -0,0 +1,20 @@
...
+    required: true
+  match_password:
+    description: 'Match password, ideally in secrets.MATCH_PASSWORD'
+    required: false
+runs:
+  using: 'composite'
+  steps:
+    - name: 'Run fastlane'
+      uses: actions/checkout@v4
+    - run: fastlane ${{ inputs.lane }}
```
moyerr: This step name looks misplaced 😛  - it will apply to the checkout step, rather than the run script step. 

Though, more to the point, these actions will run on the runner specified by the workflow job that calls them, sharing data and context from previous step. So checking out the repo may not be necessary here since it is handled in the job that invokes this action.

.github/workflows/on_release_push.yml
```diff
@@ -0,0 +1,25 @@
+name: 'On release push'
+on:
+  push:
+    branches:
+      - 'release.**'
+jobs:
+  on_release_push:
+    runs-on: macos-14
```
moyerr, line 8: FYI [`macos-latest` now resolves to `macos-14`](https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/)

.github/workflows/on_release_push.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: 'test_unit_ui'
+        uses: ./.github/workflows/actions/run_lane
+        with:
+          lane: 'test_unit_ui'
+          match_password: ${{ secrets.MATCH_PASSWORD }}
+
+      - name: 'xliff_export'
+        uses: ./.github/workflows/actions/xliff_export
+
+      - name: 'build_deploy_testflight'
```
moyerr: These step names show up in the job detail pane when the workflow is run, so I find it helpful to have human-readable step names, rather than snake-case identifiers.

Screenshot 2024-05-24 at 3 46 49 PM

moyerr (commented):
This is a great first step to migrating to GitHub Actions! I love the amount of care & thought you put into designing reusable actions with event-driven workflows. I have always relied on reusable workflows, and the limitation that a reusable workflow can only be invoked as a full-fledged job (and not as a step within a job) can be frustrating, so I may have to try your strategy soon!

I wanted to share one other tidbit that I've found helpful while iterating on my GitHub Actions. It can be annoying to need to trigger specific actions to test your workflows (e.g. opening a PR prematurely just to run it). The `workflow_dispatch` trigger allows you to invoke the workflow from within GitHub's UI on a branch of your choice. It's a huge time saver in my experience 🙂 
Screenshot 2024-05-24 at 4 06 23 PM

.github/workflows/actions/run_lane/action.yml
peterkos: Oh yes tyty -- it's the "make everything consistent", "statically-typed" brain i'm bringing over :p where I would use `"` on one line, `'` on another, `''` on another... I'll pair it down :)

.github/workflows/on_release_push.yml
peterkos: Do we always want it to point to the `latest` version though? i.e., if GitHub updates the runners, we would risk CI changing out-of-step with local dev environments

.github/workflows/on_release_push.yml
peterkos: good point! I was prioritizing "what is going on", i.e., xliff_export is "documented" in its fastlane lane, vs. splitting its description across the .workflows file and the Fastfile. Happy to add a non-snake-cased thing here tho!

.github/workflows/actions/run_lane/action.yml
peterkos: Good catch! I'll update this

.github/workflows/on_release_push.yml
moyerr: Oh yeah, definitely not requesting this change. I think either way is perfectly acceptable depending on project needs 🙂

peterkos: I'll leave it to macos-14 then 👍

peterkos:
Haven't forgotten about this ;) dub dub week's been wild, will take a look this weekend 👍

Clstroud:
Thank you, @luccab03, for picking this up and getting it back in motion! Also, thanks to @peterkos for laying the foundations here, even though I've strayed from parts of that vision in a few small ways. I think it's in a state where we can merge it and do some smaller iterative PRs to get it the rest of the way there. 

Here are the highlights of what I have adjusted:

### Test Plans
Unit and UI tests were created before Test Plans existed, and now we can use them to retry when tests fail automatically. I've set it to retry a failed test up to 3x, and it will respawn the process, hopefully sidestepping uncontrollable issues like unexpected simulator crashes.

### Crowdin Updates
I've attempted to modernize the workflow, but there are still a few unfortunate gaps I have to compensate for. Crowdin has added many new tooling and workflow options, but the output files require additional processing.

* The old CrowdinExport/CrowdinImports folders have been removed and no longer need to be managed manually
* Xcode String Catalogs can be uploaded directly to Crowdin, so there is no apparent need to export as `xliff` files anymore, removing an entire `xcodebuild` operation.
* Crowdin allows us to export a preconfigured "bundle" of xliff files corresponding to each source string catalog. We then download that bundle and import each xliff file one by one using `xcodebuild`.
* Each xliff file is almost correct, but two metadata tags are swapped, making `xcodebuild` unable to import them correctly. I do a quick regex manipulation in Ruby and then attempt to import the file.
* Xcode can struggle with the import operation due to a known issue with package resolution. This issue is typically resolved by reattempting the import, so I added a quick Ruby function to repeat the operation up to 3x. So far, that has allowed all files to be imported without issue!
* Importing all of those xliff files separately is expensive, so merging them into a single file and doing a single import may be an opportunity to optimize. However, I have not verified that this works!

### Workflow
This is subject to change with future iterations, but the overall translation workflow is mostly the same:
1. New strings are merged into `develop`, at which point they get uploaded to Crowdin.
2. Every 3 hours, a workflow runs that will fetch translations from Crowdin and attempt to import them, making a PR if anything has changed.

### TODO
1. Implement the release workflow, so we can cut builds to Testflight
2. Ensure test failure reporting and artifact uploads are working as expected
3. Integrate PR suggestions for SwiftLint and potentially SwiftFormat
4. Investigate xliff merging optimizations

PR #791, opened 2024-08-19
Keyboard Foundations Rewrite

### LINKED ISSUE #790: Rewrite Keyboard Implementation
Since Vocable launched, its keyboard UI has been built as a `UICollectionViewCompositionalLayout`. This code has proven difficult to maintain and not flexible enough to meet our needs. A new implementation should be built that allows us to express layouts more flexibly while also being easier to understand and maintain.

Clstroud:
closes #790 

I've worked on this in my spare moments for the last few months. It could probably use another round of refinement, but it's viable enough to review and see if we can find any outstanding gaps.

**Q: Why is it not in SwiftUI?**
**A:** I tried very hard to implement this in SwiftUI in at least three attempts, but the bridging required between UIKit and SwiftUI made it a nightmare. I could get close, but never all of the way there. I'm more than happy to discuss those challenges in another medium, but they are why I chose a custom DSL + UIKit. When this part of the codebase gets touched again, Vocable should be iOS 17+ only, which should ease much of that pain.

Here's an imaginary keypad-specific layout to illustrate:
```swift
import Foundation
import SwiftUI
import Collections

struct NumpadKeyboardLayout: KeyboardLayout {

    let identifier = "numpad_layout"

    func makeLayout(
        configuration: KeyboardLayoutConfiguration
    ) -> KeyboardBody {

        // Keypad rows
        let values = "123456789"
        for row in values.chunks(ofCount: 3) {
            KeyboardLayoutRow {
                for key in row {
                    KeyboardLayoutKey(key)
                        .keyWidth(count: 3)
                }
            }
        }

        // Decimal/zero/backspace
        KeyboardLayoutRow {
            KeyboardLayoutKey(".")
            KeyboardLayoutKey("0")
            KeyboardLayoutKey(.backspace)
        }
        .keyWidth(count: 3)

        // Mode toggle, Space, Speak
        KeyboardLayoutRow {
            KeyboardLayoutKey(
                configuration.mode == .alphabetical ? .numberPad : .alphabet
            )
            .keyWidth(span: 3)
            KeyboardLayoutKey(.space)
                .keyWidth(span: 6)
            KeyboardLayoutKey(.speak)
                .keyWidth(span: 3)
        }
        .keyWidth(count: 12)
    }
}

#Preview {
    KeyboardLayoutPreviewView(
        layout: NumpadKeyboardLayout(),
        mode: .alphabetical
    )
}
```

Which produces this:

Previous layouts look roughly the same but with additional features:
| Layout | Portrait    | Landscape |
| :---: | :--------: | :-------: |
| Compact |   |     |
| QWERTY |  |      |

### Next Steps
* The Settings UI for keyboard layout needs to be removed, with the "Compact QWERTY" setting being tucked under the head tracking toggle section. This reflects the new default behavior of having the alphabetically-ordered keyboard only the default in portrait mode when head tracking is enabled.


stevefosterwta (approved):
👍

SashankPatel:
Test Notes 8/27: 
Feel free to ignore some of these if they dont apply yet, I know there is a next step disclaimer...which we probs dont want to do but we can chat about it.

Verifying QWERTY to Alphabetical keyboard switch:
- Noticed on simulator tapping on the toggle button switches the keyboard, but the toggle button is not "active" unless you go back into the view.

https://github.com/user-attachments/assets/7ceab30d-dc20-4a41-9ff5-c43f5e6a4e79

Verified Compact QWERTY Toggle is still working ✅.
Verified when changing from landscape and the Compact Keyboard toggle is off: Alaphabetical is displayed in Portrait and QWERTY is displayed in Landscape. ✅

Question: Does the keyboard support other languages?

Clstroud:
@SashankPatel thanks for taking a look!

1. The issue in your video should be resolved by #793. I needed to draw the line somewhere and split out multiple work units, so those fixes happened as a follow-up.
2. Unless there is a bug I'm unaware of, this new keyboard is meant to implement the same variants we had before: English, Italian, and German. We can build other layouts as necessary, but those are the ones that existed previously, so I attempted to "port" them over to the new system.

SashankPatel (approved):
🚀

PR #802, opened 2025-01-03
[795] Prospective fix for audio engine crash

### LINKED ISSUE #795: [Build 1.7] - App crashes when Coming back from foreground.
Seen a couple instances where after backgrounding and foregrounding the app, it will crash.

This noteably happened in the following scneario but unable to reliably reproduce.

Steps:
1. Launch the app
2. Tap on the Settings icon
3. Tap on Voice
4. Tap on Personal Voices
5. Top on the Open Settings button.
6. Background the settings screen.
7. Tap on the Vocable icon.

Clstroud:
Attempts to resolve https://github.com/willowtreeapps/vocable-ios/issues/795

* Switch audio buffer callback mechanism to invoke observers directly instead of doing a dispatch through Combine
* Fix `EXC_BAD_ACCESS` stemming from objects captured in closure getting clobbered 
* Fix audio engine failing to initialize on iOS 17+ due to audio session not being active. <-- This is a strange one and may have UX implications. Needs validation on a physical device.


stevefosterwta (approved):
👍

Issue #826, opened 2026-02-19
Have Compatibility error on devices/os's where personal voice is not supported


SashankPatel:
**Acceptance Criteria**:

GIVEN I want to provide appropriate messaging to the user
WHEN I am on a device lower than iPhone 12 or an iOS lower than 17
THEN I should display a compatibility error noting so.

**Design**:

Screenshots or Link to Figma

SashankPatel:
Currently its not straightforward to do this because iOS's personalvoice api does not provide information on whether a device is supported or not.  For now going to say this is not be needed as there are only a handful of users using unsupported devices.

PR #827, opened 2026-02-20
BugFix - GH-826: Add compatibility error for personal voice.

SashankPatel:
https://github.com/willowtreeapps/vocable-ios/issues/826

closes 826

# Description of Work
Add a compatibility error for personal voice on devices < iOS 17 and < iphone 12.  Also removed teh pagination footer.

| Before | After |
|--------|--------|
| IMG_0144| Screenshot 2026-02-20 at 8 22 03 AM|

## Notes to Test (Optional)
Notes

---

### DIFF
--- a/Vocable/AppConfig/AppConfig.swift
+++ b/Vocable/AppConfig/AppConfig.swift
@@ -39,6 +39,39 @@ struct AppConfig {
         return ARFaceTrackingConfiguration.isSupported
     }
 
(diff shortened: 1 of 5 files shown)


stevefosterwta (approved):
👍

SashankPatel:
Since Apples doesn't give a clean way in the personal voice API to detect supported devices and according to mixpanel there are only a hand full of users using unsupported devices.  Instead we will keep the original implementation and update the language saying personal voice is only supported on certain devices.  Will create a new pr with the changes to keep the PR clean.

